PR #12, opened 2015-12-14
Windows Icon: Try a white version for Windows 10.

jasmussen:
This PR adds a white version of the app icon, to make it look more at home on Windows 10.

Can we keep the colored version (kept in this PR) and serve that to Windows 7 users? Or does that require a separate build of the app?

The downside to a white icon is that it's invisible in the file explorer:

white-icon

I'd say this is acceptable though, as I don't think it's that often you go looking at the .exe files. 

Here's how the icon looks in your taskbar:

taskbar-win10

taskbar-win7

My opinion: if we can serve a white icon for Windows 10, and a colored icon for Windows 7, without creating separate builds and doing an inordinate amount of work, I think this is a good change.


AdamHTML7:
Yes, this would be an excellent solution because the application in invisible on the file explorer like he said.

mkaz:
Tested it out and still does not work as hoped. The install script can detect the version but there is an additional icon specified in the main build resource config, which can not detected during install but only on build. 

Looking at OS usages, <a href="https://www.netmarketshare.com/operating-system-market-share.aspx?qprid=10&qpcustomd=0&qpcustomb=">Win7 still has a large majority of install </a> - so best bet is to probably wait and switch icon when Win10 has majority.

Also, there is some unintended consequences when switching the logo, so will need to be careful when it comes to this.

screen shot 2016-05-31 at 11 32 33 am

jasmussen:
Bummer. This suggests to me we should still keep the full colour logo as the only one then. I suppose we can revisit this ticket if things change. Thanks for testing.

AdamHTML7:
I agree with @jasmussen that we should keep the full color logo. Maybe try to create a white/black one that looks good on any desktop/dock?

astralbodies:
I'm closing this PR for now given the age; feel free to re-address if it's still valid.

PR #45, opened 2016-01-20
Boot Calypso in a forked child_process

mjangda:
We do this by running the app bundle directly, instead of running the server directly within the Electron app.

This lets us run the transpiled server code and enjoy ES6.


desktop/server/server.js
```diff
@@ -25,18 +25,47 @@ function showFailure( app ) {
...
+			startedCallback();
+		}
+		debug( 'app.stout: ' + data );
+	} );
+
+	app.stderr.on( 'data', function( data ) {
+		debug( 'app.stderr: ' + data );
+	} );
+
+	app.on( 'close', function( code ) {
```
gwwar: Maybe also check for signal here:
```
function( code, signal ) { ... } 
```
Currently sending an interrupt results in:
```
desktop:server app.close with code: null +5ms
```

desktop/server/server.js
```diff
@@ -25,18 +25,47 @@ function showFailure( app ) {
...
+		silent: true,
+		cwd: path.resolve( process.cwd(), 'calypso' ), // Force working directory to calypso root
+		env: env
+	} );
+
+	app.stdout.on( 'data', function( data ) {
+		// We need to wait until the server has booted.
+		// Let's use the "booted" message as a signal that it's ready.
+		var str = data.toString();
+		if ( str.indexOf( startedSignal ) === 0 ) {
```
mjangda: This is a bit annoying and fragile. If Calypso ever changes the wording of the boot message, the UI will fail to load.

mjangda: Options:

- Add a way for Calypso to send us a proper signal on boot.
- Keep watching for the boot message but add a `setTimeout` with `startedCallback` as a fallback.
- `setTimeout` with retries (+ backoff) using `portScanner`.

gwwar: Children are able to send messages back to the parent process: https://nodejs.org/api/child_process.html#child_process_child_send_message_sendhandle_callback

So it should look a bit like this in calypso:
```
process.send({ boot: 'ready' });
```
And in this file:
```
app.on('message', (m) => {
  console.log('PARENT got message:', m);
});
```

gwwar:
:+1: Thanks for working on this! I tested this against the bundle version, along with calypso master and it works great.

I'm not sure I follow why forking a child fixes this. Were we previously using some calypso files directly in addition to the desktop bundle?

desktop/server/server.js
mjangda: Nice! `process.send` is only available for forked processes (e.g. would not be when running running Calypso as-is), so thinking of something like this:

```
diff --git a/index.js b/index.js
index 02ff4d8..dedba3a 100644
--- a/index.js
+++ b/index.js
@@ -23,6 +23,10 @@ console.info( chalk.cyan( '\nGetting bundles ready, hold on...' ) );
 server = http.createServer( app );
 server.listen( port );

+if ( process.env.CALYPSO_IS_FORK ) {
+       process.send( { boot: 'ready' } );
+}
+
 // Enable hot reloader in development
 if ( config( 'env' ) === 'development' ) {
        hotReloader = require( 'bundler/hot-reloader' );
```

Desktop can set the `env` var when forking + watch for the `boot` message.

gwwar: :+1: Looks like a good strategy!

gwwar:
There's a few cases where the parent process can be killed without killing the child. (For example try running the app without your proxy on). We might need to call child.kill when the parent process exits.

desktop/server/server.js
mjangda: Calypso change: https://github.com/Automattic/wp-calypso/pull/2628

mjangda:
> I'm not sure I follow why forking a child fixes this. Were we previously using some calypso files directly in addition to the desktop bundle?

Prior to this, we were just booting Calypso directly within the desktop process by require-ing raw Calypso code, which is why we ran into node compatibility issues with ES6.

If we `require` the transpiled build directly (`require( 'path/to/bundle-desktop.js' )`), it has issues due to the working directory (which is set to the desktop app root) and fails to resolve config files and such. I'm sure we could have fixed this but it feels cleaner to run desktop and calypso in separate processes (admittedly, it adds complexities like your note about watching for when the parent dies).

gwwar:
Thanks for the explanation! I think this would be useful to add in the development.md readme.

mjangda:
81d459e updates docs.

mjangda:
> There's a few cases where the parent process can be killed without killing the child.

Yeah, looks like easiest way to reproduce is to start the app, and then quit with Cmd+Q. Electron is killed but the Calypso server is still running in the background (according to `ps`).

mjangda:
Works well via `make` but there are some issues with the packaged versions (server fails to start; looks related to path resolution). Looking into it.

johngodley:
Nice work Mo! I also like the branch name :smile: 

I've tried a few scenarios and it seems to handle most things, even if an error is thrown in the server when a request is made.

One possible issue is if Electron itself crashes and the server process is then not killed (you can `kill -9` on the main process to simulate this). I don't think there's much we can do to handle the crash when it happens, but I wonder if we can recover afterwards.

For example, we detect a server is already running when we start the app. If it is, maybe we can somehow re-attach rather than start. Alternatively, maybe the server process can occasionally ping the parent process - if it doesn't respond it kills itself.

It's more of an edge case, but if Electron ever does crash then it would require a reboot to get the app working again.

mjangda:
> If it is, maybe we can somehow re-attach rather than start. Alternatively, maybe the server process can occasionally ping the parent process - if it doesn't respond it kills itself.

This sounds like a good approach.

mjangda:
I wonder if we should merge this as-is for now and sort out the packaging issues in a separate PR especially since it's potentially blocking people.

johngodley:
:+1:

PR #49, opened 2016-01-24
Passing command line arguments to desktop app

artpi:
This is a desktop part of an effort to bring command line argument passing to desktop app.
Here is complimentary PR:
https://github.com/Automattic/wp-calypso/pull/2727

### DIFF
--- a/desktop/lib/app-instance/index.js
+++ b/desktop/lib/app-instance/index.js
@@ -5,6 +5,10 @@
  */
 const app = require( 'electron' ).app;
 const debug = require( 'debug' )( 'desktop:app-instance' );
+const minimist = require( 'minimist' );
+const fs = require( 'fs' );
+const path = require( 'path' );
+
 
 /**
  * Internal dependencies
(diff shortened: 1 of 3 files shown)


johngodley:
Can you describe how this will be used?

When I looked at OS X services before it seemed like it would be best for the app to add itself to the services directly by using a native module.

artpi:
It was my intention to keep discussion brief in this PR - the explanation, why and a use case is in the linked WP-calypso PR. And yes, services would be perfect :)

johngodley:
Sure, I understand not wanting to duplicate info across PRs. However, I'm trying to find more out about how this specific PR would be used in the context of the desktop app, so it seemed appropriate to comment here.

For example, you say it would be for services - how would parsing command line arguments be used to do this? That is, can you show how this PR would be used in a flow that provides services. The code itself looks fine, I just don't understand the bigger picture of how it would be used.

artpi:
You can create an Automator service that adds a service to services menu but underneath passes command line arguments. We could package this Automator service with the desktop app. Apart from services just beeing able to read post body from a text file is also useful in my opinion

johngodley:
Gotcha. If we can bundle the Automator script with the app and have the app be able to run and install the service then this sounds like a good way of adding the functionality.

I'd probably go with something more descriptive than `args` though - maybe `command-line-arguments` etc

artpi:
@johngodley : Automator script in action: https://github.com/Automattic/wp-calypso/pull/2727#issue-128433291

PR #54, opened 2016-01-26
Improve build script for Mac App Store

gziolo:
When you try to run Mac App Store version of desktop app it uses default desktop config. This PR fixes that. Additionally it changes the Calypso build script name to allow easy switch between different Desktop app versions: standalone vs Mac App Store. 

This change depends on this change Automattic/wp-calypso/pull/2781. They should be merged with master at the same time.

### Testing
Apply Calypso changes from Automattic/wp-calypso/pull/2781.

1. Execute `make run`.
2. Desktop app should launch with default desktop config.
3. Close app.
4. Execute `make run-mas`.
5. Desktop app should launch with desktop config adjusted for Mac App Store.
6. Close app.
7. Repeat (1) and (2). You should see default app setup again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ CERT_SPC := $(THIS_DIR)/resource/secrets/automattic-code.spc
 CERT_PVK := $(THIS_DIR)/resource/secrets/automattic-code.pvk
 CALYPSO_DIR := $(THIS_DIR)/calypso
 CALYPSO_JS_STD := $(CALYPSO_DIR)/public/build-desktop.js
-CALYPSO_JS_MAS := $(CALYPSO_DIR)/public/build-desktop.js
+CALYPSO_JS_MAS := $(CALYPSO_DIR)/public/build-desktop-mac-app-store.js
 CALYPSO_CHANGES_STD := `find "$(CALYPSO_DIR)" -newer "$(CALYPSO_JS_STD)" \( -name "*.js" -o -name "*.jsx" -o -name "*.json" -o -name "*.scss" \) -type f -print -quit | grep -v .min. | wc -l`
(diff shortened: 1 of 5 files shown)


johngodley:
Looks good. This is what I intended to do but never got round to it so thanks for finishing it off!

Are there any differences in the Calypos server? The desktop app currently uses `calypso/build/bundle-desktop.js` and I note there is a `calypso/build/bundle-desktop-mac-app-store.js` (see `desktop/server/server.js`)

We can use the config `calypso_config` setting instead as this has the correct value to use.

Also take a look at `env.js` on line 25 - there is a todo mentioning that we may have the wrong CALYPSO_ENV value set. Your PR means we are now using the correct JS file so the todo can be removed. I'm not sure if CALYPSO_ENV should always be `desktop`, or whether it needs to be `desktop-mac-app-store`.

gziolo:
> We can use the config calypso_config setting instead as this has the correct value to use.

Good catch. I'm not sure if there are any differences, but we definitely should update that part.

> Also take a look at env.js on line 25 - there is a todo mentioning that we may have the wrong CALYPSO_ENV value set. Your PR means we are now using the correct JS file so the todo can be removed. I'm not sure if CALYPSO_ENV should always be desktop, or whether it needs to be desktop-mac-app-store.

My guess is that we need to keep it this way to load properly Jade file: https://github.com/Automattic/wp-calypso/blob/master/server/pages/desktop.jade. Another approach would be creating 2 files on Calypso side: 
- desktop.jade
- desktip-mac-app-store.jade
that include common Jade file and server different script name.

gziolo:
@johngodley I added 2 fixes based on your comments.

It turned out that `desktop.jade` display is based on `desktop` feature flag. See https://github.com/Automattic/wp-calypso/blob/c7ef9c00e2438ad0cef90a3530e8fefbf103f07f/server/pages/index.js#L312. I did some testing and it looks like we can remove todo and use env set in config file :)

johngodley:
LGTM

PR #56, opened 2016-01-27
Disable scroll bounce

artpi:
# What are you talking about?

OSX has natively this effect of 'bounce' when you are scrolling to the edge of current window.
I HATE THIS in the app.

# Y u no CSS?
It is possible to achieve the same effect in CSS:
```css
html.is-desktop {
overflow:hidden;
}
html.is-desktop body {
overflow: auto;
}
```

But this hides the scrollbar under masterbar and also messes with scrollbar visibility (the scrollbar disappers, but scrolling is still working)

# Better solution
We discussed it with @rralian and decided that even better solution would be to play this video when while scrolling, we
**[push it to the limit / past the point of no return](https://youtu.be/kZu5iDTtNg0?t=24s)**
Unfortunately, this clip is not GPL :/

# Testing
Just scroll with swipe to the edge

CC @adambbecker @johngodley @dmsnell @gziolo

### DIFF
--- a/public_desktop/desktop-app.js
+++ b/public_desktop/desktop-app.js
@@ -90,6 +90,15 @@ function startDesktopApp() {
 		}
 	}
 
+	function preventScrollBounceOSX( e ) {
+		if (
+			( e.deltaY < 0 && document.body.scrollTop === 0 ) ||
+			( e.deltaY > 0 && document.body.scrollTop === document.body.scrollHeight - window.innerHeight )
+		) {
+			e.preventDefault()
+		}
+	}
+
 	debug = gGebug( 'desktop:browser' );
 
 	// Everything is ready, start Calypso
@@ -102,6 +111,10 @@ function startDesktopApp() {
 
 		document.addEventListener( 'keydown', keyboardHandler );
 		document.addEventListener( 'click', preventNewWindow );
+
+		if ( window.navigator.userAgent.indexOf( 'Macintosh' ) !== -1 ) {
+			document.body.addEventListener( 'mousewheel', preventScrollBounceOSX );
+		}
 	}
 
 	// This is called by Calypso


johngodley:
Apart from a missing semicolon on the `preventDefault` it looks good to me.

Pinging @jasmussen for thoughts.

jasmussen:
I also hate this rubberbanding scroll effect. I'm also glad you didn't go a CSS route, because we've already had issues tweaking the overflow. However I'm not convinced this should be an "app-only" thing, for a number of reasons.

Most importantly, every time we do a change that's app-only, we fragment our efforts between wp-calypso and wp-desktop, and actually cause technical debt we need to remember. 

While there are many things we can and should do to make the app "feel" more native than it actually is,  I feel like we're disregarding the legacy of the web, and the fact it's a web-app. For example on the web, clickable things have a hand cursor, in apps they have the default cursor. But this is not a native app, and any changes we make to pretend it is, I fear, will just cause confusion, and ultimately fragment the actual usability of this app compared to the web-app. 

Bottomline is, on key areas like this I feel like the desktop app should mimic the web-app 1:1. And so if you'd like to disable the rubber banding effect, I feel like it should be done in an upstream wp-calypso PR, and not here. If you all strongly disagree with my assessment here, at the very least it should still be code that lives in Calypso and is then enabled in the desktop app through a feature flag.

jasmussen:
CC: @mtias

artpi:
I am closing this PR.
Per @jasmussen advice, it is now in wp-calypso behing a featur flag: https://github.com/Automattic/wp-calypso/pull/2929

onlinehandyman:
UNSUBSCRIBE

onlinehandyman:
Please UNSUBSCRIBE from list

On Sat, Jan 30, 2016 at 5:14 PM, Artur Piszek <notifications@github.com>
wrote:

> Closed #56 <https://github.com/Automattic/wp-desktop/pull/56>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Automattic/wp-desktop/pull/56#event-532673783>.
>



-- 
*Visit Us At:*

http:// <http://shelf-concepts.com>onlinehandymanwebdesign.com

mtias:
@jasmussen sorry for the delay, I completely agree with your reasoning. My concern though is that hijacking the default OS behaviour may seem rather odd in a browser. Regardless, having this in wp-calypso anyways makes total sense, even if we enable it only for desktop via feature flag.

PR #60, opened 2016-01-29
Update build scripts to prune more

johngodley:
This updates the build scripts so they:

- Ignore most of the Calypso server and client source files (we use bundled files)
- Use a specific subset of node modules for the Calypso server

This is a little brittle, but reduces the number of packaged files by about 12k.

A better solution is to investigate what the Calypso is doing in the asset bundler that requires external modules, and if it's possible to drop this in the app. Doing so may also mean we can enable asar again.


mjangda:
:boom:

PR #64, opened 2016-02-04
Ping Pong!

### LINKED ISSUE #51: Calypso server persists after Electron crashes
From #45 via @johngodley:

This is an edge case: if Electron crashes, it requires a reboot to get the app working again because the server process is then not killed (you can kill -9 on the main process to simulate this). There isn't much we can do to handle the crash when it happens, but we can try to recover afterwards.

For example, we detect a server is already running when we start the app. If it is, maybe we can somehow re-attach rather than start. Alternatively, maybe the server process can occasionally ping the parent process - if it doesn't respond it kills itself.

mjangda:
Make the app send periodic pings to Calypso to make sure the server is still running. If we fail to get a response, we can detect that and handle accordingly.

Needs https://github.com/Automattic/wp-calypso/pull/3062 to work

Fixes #51

### DIFF
--- a/desktop/server/server.js
+++ b/desktop/server/server.js
@@ -17,10 +17,13 @@ const Config = require( '../config' );
  * Module variables
  */
 const BOOT_TIMEOUT = 10000;
+const PING_TIMEOUT = 10000;
 const ERROR_TIMEOUT = 1;
 const ERROR_EXITEARLY = 2;
+const ERROR_SERVERPINGTIMEOUT = 3;
 
 let didBoot = false;
+let pingKillswitch = null;
 let calypso = null;
 
 function showFailure( app, error, code ) {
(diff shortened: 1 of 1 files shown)


mjangda:
We need to decide if `maybeKillswitch` should kill the calypso server or restart it.

Also, looks like there may be issues when going into screensaver or sleep mode.

johngodley:
Seems to work although I've not been able to get the `Failed to get ping from Calypso; exiting.` message to appear (I may be doing something wrong, or the app exits before the message is shown). No processes are left behind.

What happens when the system goes to sleep - does the kill timer expire on wakeup?

mjangda:
I can't reproduce the sleep issue now. I think it may have been something else.

Things I've tested:

### Kill Electron

- `make run`
- find pids using `ps aux | grep [e]lectron`
- `kill -9` the 2 main electron processes using the pids
- `ps aux | grep [e]lectron` shortly after and verify the `bundle-desktop.js` process is no longer running

### Kill Calypso

- `make run`
- find pid for `bundle-desktop.js` using `ps aux | grep [e]lectron`
- `kill -9` using the pid
- console shows the following:

```
desktop:server app.exit with code: 0 (SIGKILL) +2s
  desktop:server app.close with code: 0 (SIGKILL) +1ms
  desktop:server Failed to get ping from Calypso; exiting. +28s
  desktop:server Killing server +1ms
```

The window stays open and the app continues to mostly work (since it's an SPA :)) but breaks if you log out. In this instance, it would be nice if we could gracefully restart the server, or show an error to the user.

mjangda:
No longer necessary as of #66

PR #65, opened 2016-02-04
Track user's last location

mjangda:
Store the user's last location in settings and navigate to that when they re-open the app.

## To Test

- Start app
- Navigate somewhere
- Exit app
- Restart app
- Ensure the app opens at your same location you exited

/cc @jasmussen

### DIFF
--- a/desktop/lib/settings/constants.js
+++ b/desktop/lib/settings/constants.js
@@ -1,0 +1,3 @@
+module.exports = {
+	LAST_LOCATION: 'last-location'
+}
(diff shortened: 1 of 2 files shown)


mjangda:
I guess one thing to be careful of is when a path is removed. Let's say I close the app at `/me` and then update at some point and we've changed that path to `/my-wpcom`, we should make sure that's handled gracefully.

That feels like something Calypso would generally do but probably still good to have a failsafe.

mjangda:
One pretty nice touch with this: if you are writing a post and quit, the app will re-open to that post.

johngodley:
Looks good.

Also, furthering your broken page test one thing I could see happening is the user leaves the app on a page that refers to a particular resource (a site, for example). They then delete that resource from the web and restart the app. As you say, something that should be handled in Calypso.

PR #66, opened 2016-02-04
Webpack app and server

johngodley:
This is a different route to getting the app working with the ES6 Calypso server. It uses webpack to bundle the desktop app code with the Calypso server, producing a single bundled file:

- Server runs in same process, so no problems with sandboxing or orphaned processes
- One file for whole app (excepting node modules for Express and Superagent)
- ES6 in Calypso is fine (it goes through Babel before running)
- A `build` directory is used to create a temporary working copy of the app for both dev and production. This should make it easier to control what goes into the app (it whitelists files rather than blacklisting what shouldnt be included), and should make the dev build behave more like the production build

I've tested this running in dev and packaged modes on OS X.

Additional comments:
- Uses minified Calypso client

Things still to do:
- Improve Babel usage so app can fully use ES6. Currently when I try to add the `es2015` preset to Babel it throws lots of errors
- Get asar working - it throws an error
- Minify the bundle - it throws errors
- Improve the `make package` rule
- ~~Get unit test working again~~

Disadvantages
- Need to rebuild the bundle each time you run the app in dev mode


mjangda:
Looks like the `asar` fails at this: `process.chdir(app.getAppPath())`

johngodley:
> Calypso fails in config because the paths don't resolve properly in the asar

Yeah, when the app runs in dev mode the `process.chdir` isn't needed - the `node: { __dirname: true }` webpack config takes care of the path. When packaged the `__dirname` value is set to `/`, which from what I've read on Electron is expected behaviour.

Currently I'm thinking we either ignore asar, or modify Calypso so that we can supply a base path and stop it using `__dirname`.

Issue #83, opened 2016-02-11
Windows: Delete AppData directory on uninstall


mkaz:
The AppData directory `C:\Users\<username>\AppData\Roaming\WordPress.com` persists after an uninstall. We had an instance that this was corrupted and required manually deleting the directory to fix the issue. Uninstalling the app did not help.

Need to investigate:
* Will the uninstall script know the username?
* What about multiple users on the system?

mkaz:
Issue fixed in PR #86. Note for testing, by deleting the AppData directory, you will be logged out after an uninstall since user data is removed.

PR #86, opened 2016-02-15
Add RMDir to uninstall to remove AppData

mkaz:
Remove AppData\Roaming\WordPress.com directory on uninstall. This fixes
the issue with corrupt user data, or staying logged in after an
uninstall. Properly cleans up after itself.


mkaz:
Fixes #83

PR #103, opened 2016-02-25
Fix #99 - Add try-catch around loading settings

### LINKED ISSUE #99: SettingsFile.load fails for invalid files
We should better handle cases where the settings-file gets corrupted and not crash the app.

```
SyntaxError: Unexpected end of input
at Object.parse (native)
at Object.module.exports.load (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:25815:17)
at Settings._getAll (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:2883:33)
at Settings.isDebug (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:2890:19)
at Object. (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:77102:15)
at __webpack_require__ (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:21:30)
at Object. (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:77027:2)
at __webpack_require__ (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:21:30)
at Object. (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:51:2)
at __webpack_require__ (/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:21:30)
```

mkaz:
If the settings file is corrupt than it will recreate a new settings
file, same as if the settings file didn't exist in the first place.

### DIFF
--- a/desktop/lib/settings/settings-file.js
+++ b/desktop/lib/settings/settings-file.js
@@ -6,6 +6,7 @@
 const app = require( 'electron' ).app;
 const path = require( 'path' );
 const fs = require( 'fs' );
+const debug = require( 'debug' )( 'desktop:settings' );
 
 /**
  * Internal dependencies
@@ -28,7 +29,12 @@ module.exports = {
 		const settingsFile = getSettingsFile();
 
 		if ( fs.existsSync( settingsFile ) ) {
-			return JSON.parse( fs.readFileSync( settingsFile ) );
+			try {
+				return JSON.parse( fs.readFileSync( settingsFile ) );
+			}
+			catch (e) {
+				debug( 'Error reading settings file' );
+			}
 		}
 
 		firstRun = true;
@@ -37,7 +43,7 @@ module.exports = {
 	},
 
 	save: function( group, groupData ) {
-		const debug = require( 'debug' )( 'desktop:settings' );
+
 		const settingsFile = getSettingsFile();
 		let data = {};
 


desktop/lib/settings/settings-file.js
```diff
@@ -28,7 +29,12 @@ module.exports = {
 		const settingsFile = getSettingsFile();
 
 		if ( fs.existsSync( settingsFile ) ) {
-			return JSON.parse( fs.readFileSync( settingsFile ) );
+			try {
+				return JSON.parse( fs.readFileSync( settingsFile ) );
+			}
+			catch (e) {
+				debug( 'Error reading settings file' );
```
mjangda: Should we do something else here? Because we'll fall through and the app will think it's a first run (even though it isn't) and try to create a the settings file which could be problematic.

mkaz:
I tested it out - if the file exists but can't be read than it is likely corrupt, which means it not really worth keeping around, so creating a new one shouldn't be a problem.

I tested this out by putting a corrupt settings file in place, opening the app with this change and it clears it out with a valid one. 

Here's what currently is in a settings file, nothing critical if lost. The real issue would be to notify the user settings was corrupt and replaced with defaults, could do in an alert in the "catch"

```
{
	"notification-badge": true,
	"notification-bounce": false,
	"release-channel": "stable",
	"proxy-type": "",
	"proxy-url": "socks5://localhost",
	"proxy-port": "8080",
	"proxy-pac": "",
	"last-location": ""
}
```

PR #105, opened 2016-02-26
Implement Basic i18n Support

mjangda:
This PR has a few pieces:
- Script to fetch translation files during build process.
- Load up translation when system locale doesn't match the default (`en`) and we have a translation file available.
- Use Calypso's i18n mixin to translate text in the Menus.
## To Test
- Run `make fetch-translations` to fetch translation files
- Change your system language to something non-English like Spanish
- Run the app using `make run`.
- Since we don't have translations for a lot of things, many things will still show in English. Most of the Calypso shortcuts in the "Window" menu item. should be translated at least.
- Make sure the client itself matches your Calypso setting.

Then:
- Switch back to English.
- Run the app again using `make run`.
- Make sure everything is in English.
## Future

Other improvements/TODOs for future PRs:
- Implement translations for other parts of the app like Preferences, etc.
- Pass the local translation data through to the client to avoid the remote request.
- Maybe handle mismatch between system and Calypso language settings.

See #8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ publish/publish-config.js
 publish/node_modules/
 resource/secrets
 build
+public_desktop/languages/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ publish/node_modules/
 resource/secrets
 build
 public_desktop/languages/
+wpcom-desktop-strings.php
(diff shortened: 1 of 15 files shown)


astralbodies:
@mjangda This PR is really old and I'm not sure what happened to it's original review. Is this something we want to get updated and try reviewing again?

mjangda:
We never finished reviewing/launching i18n support. Probably needs a major refresh, especially to catch up with any major changes on the Calypso side. Some bits might be re-usable but might be easier just to start from scratch.

astralbodies:
Okay thanks for that bit of insight! I'll close this PR for now but will leave the branch alone.

PR #111, opened 2016-03-03
Add isValidLastLocation() to check 

### LINKED ISSUE #110: Don't save or use "bad" last locations
The last location can get cached at a bad location such as `/desktop/hey.html` which gets displayed if Electron is not available. However, next time app is launched it will stay there.

This likely can happen for other error or similar pages we don't want to always go back to.

mkaz:
Creates the `isValidLastLocation()` function to use when navigating to and attempting to store the last location. We do not want to cache an error page or start back in the middle of the sign up flow.

Fix #110

### DIFF
--- a/desktop/server/index.js
+++ b/desktop/server/index.js
@@ -77,8 +77,8 @@ function startApp( started_cb ) {
 
 function isValidLastLocation( loc ) {
 	const invalids = [
-		'/desktop/hey.html',	// Page shown when no Electron
-		'/start'				// Don't attempt to resume the signup flow
+		'/desktop/',     // Page shown when no Electron
+		'/start'         // Don't attempt to resume the signup flow
 	];
 
 	for ( let s of invalids ) {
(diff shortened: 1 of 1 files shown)


desktop/server/index.js
```diff
@@ -75,6 +75,21 @@ function startApp( started_cb ) {
 	} );
 }
 
+function isValidLastLocation( loc ) {
+	const invalids = [
+		'/desktop/hey.html',	// Page shown when no Electron
```
mjangda: Should we just block all entries in `/desktop/`?

desktop/server/index.js
```diff
@@ -75,6 +75,21 @@ function startApp( started_cb ) {
...
+		'/start'				// Don't attempt to resume the signup flow
+	];
+
+	for ( let s of invalids ) {
```
mjangda: Can probably use the `every` helper:

```
return invalids.every( function( value ) {
    return ! loc.startsWith( value );
} );
```

mjangda:
Two suggestions, but :+1:

desktop/server/index.js
mkaz: Yeh, doesn't look like any of it should be stored, since most is just menus.

desktop/server/index.js
mkaz: A bit too tricky for my taste. ;-)
Nice and concise but hard to mentally parse what will get returned and why

mjangda: Fair enough :)

PR #112, opened 2016-03-03
Fix/108 remove mas

### LINKED ISSUE #108: Remove MAS Support
We do not plan to create a Mac App Store release.
Remove the build scripts and packages that were included to support a MAS release.

mkaz:
We do not plan on creating a Mac App Store build, so this change removes the bits and pieces in the build system setup for that. Simplifies the number of targets in the Makefile, allows us to use the standard electron-packager and cleans up code that is not used.

Fix #108


package.json
```diff
@@ -28,31 +28,34 @@
...
+    "babel-preset-es2015": "^6.6.0",
     "debug": "^2.2.0",
     "express": "^4.13.3",
     "jade": "^1.11.0",
     "lodash.debounce": "^4.0.0",
     "portscanner": "^1.0.0",
-    "spellchecker": "^3.1.3",
-    "superagent": "^1.4.0"
+    "spellchecker": "^3.2.3",
+    "superagent": "^1.7.2"
```
mjangda: There's a lot of dependencies changing here which seems like it's not intentional.

mjangda:
We should revert all the unrelated version bumps. Everything else looks good!

mkaz:
Yeh, it was bit unintentional, as part of the `npm update` trying to figure out what was wrong with the build. We can do the package update in a proper PR for it. Also, I think we need to move away from using `^ versions` since it does not give consistent builds between developers. 

What npm install for me might be different than what it installs for you depending on when we run an `npm install` and what changes upstream

PR #115, opened 2016-03-07
Fix #113 - On login page, redirect to en.support.wordpress.com

### LINKED ISSUE #113: Help Menu on Login screen
The new help menu does not work if not logged in.

It requires being in Calypso, so does not work when used on the login screen. 
I suggest to update to trigger opening: https://en.support.wordpress.com/

mkaz:
The new help page within Calypso requires a user to be logged in, which
is obviously not the case when on the /login page. Adds check in Help
Menu if on /login to redirect to external support page.

### DIFF
--- a/desktop/lib/menu/help-menu.js
+++ b/desktop/lib/menu/help-menu.js
@@ -5,7 +5,6 @@
  */
 const shell = require( 'electron' ).shell;
 const ipc = require( 'lib/calypso-commands' );
-const url = require( 'url' );
 
 /**
  * Internal dependencies
--- a/desktop/lib/menu/help-menu.js
+++ b/desktop/lib/menu/help-menu.js
@@ -12,6 +12,7 @@ const url = require( 'url' );
  */
 const platform = require( 'lib/platform' );
 const WindowManager = require( 'lib/window-manager' );
+const state = require( 'lib/state' );
 
 let menuItems = [];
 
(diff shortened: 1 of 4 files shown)


desktop/lib/menu/help-menu.js
```diff
@@ -30,8 +31,14 @@ module.exports = function( mainWindow ) {
 		{
 			label: 'How can we help?',
 			click: function() {
-				mainWindow.show();
-				ipc.showHelp( mainWindow );
+				// on login page - user logged out
+				let parsedURL = url.parse( mainWindow.webContents.getURL() );
+				if ( parsedURL.pathname === '/login' ) {
```
mjangda: We have more routes now, with Sign Up in the app so this doesn't cover all cases. It would be better to check if the user is currently logged in, instead.

mkaz:
@mjangda Agreed. Reworked to have a State module which can be used for logged in & out - and easily extendable when we want

desktop/lib/menu/help-menu.js
```diff
@@ -5,12 +5,14 @@
  */
 const shell = require( 'electron' ).shell;
 const ipc = require( 'lib/calypso-commands' );
+const url = require( 'url' );
```
mjangda: Can remove this now.

mjangda:
Nice, one small thing and :ship:

Issue #101, opened 2016-02-24
Spell Checking is broken for languages other than English (OSX 1.2.7)


arturogoga:
Version 1.2.7 introduced spell checking, with no apparent way to turn the feature off. I use Wordpress in Spanish, and the spell checking option seems to be completely broken for languages other than english. 

[Here](http://imgur.com/SJG2oUq) you can see how, even though the menus are shown in spanish, the Wordpress Desktop app keeps trying to spell check for English.

jasmussen:
Thanks for the report! I pulled in your image from Imgur here to Github:

![sjg2ouq](https://cloud.githubusercontent.com/assets/1204802/13289229/8b88ff92-db0f-11e5-820e-e970a73ef225.jpg)

arturogoga:
Just to let you guys know the bug is still present in the 1.3.0 Merry Weather Beta update, making the application really hard to use for people who don't post in english


![screenshot 2016-03-04 09 19 26](https://cloud.githubusercontent.com/assets/17454066/13529430/34718346-e1ea-11e5-9a8c-9523e237e9d9.jpg)

mjangda:
@arturogoga can you confirm what languages you're using for your operating system and for WordPress.com (i.e. the value at `Me > Account Settings`; from the screenshot, it looks like English).

mjangda:
It would be nice if we could detect the language for each word and toggle spellcheck errors accordingly (this is amazingly what Chrome does. I typed up a post with English, Spanish, and French words, and it didn't flag any of the non-en words as misspelled and even correct the spelling for a french word).

For an immediate fix, though, I think we can do the following:

- Don't enable spellcheck when the Calypso language is non-`en` (right now, we use the OS language).
- Add an preference to disable spellcheck (for users that prefer English as the Calypso UI language but write in other languages)

@jasmussen can I get a sanity check on this?

jasmussen:
> Don't enable spellcheck when the Calypso language is non-en (right now, we use the OS language).

:+1: 

> Add an preference to disable spellcheck (for users that prefer English as the Calypso UI language but write in other languages)

:+1: 

I think both of those sound good to me.

arturogoga:
@mjangda My Operating System is in English. I tried putting the applicaction (through Me > My Account Settings) both in Spanish and English, and the spell checking still tried to detect words in english. Which is weird because in OS X I have the system's spell checker configured in Spanish, as you can see here. 

[...]

PR #116, opened 2016-03-08
Fix #33 - Add Context Menu on Right Click

### LINKED ISSUE #33: Enhancement: Right-click support
Having right-click support to copy links (like in the Reader) would be a great addition :)

mkaz:
Add a context menu on right click within textareas, primary the Editor.

* Allow for Cut-Copy-Paste on Right Click
* Include Spelling suggestions on right-click 

Related: #101

### DIFF
--- a/desktop/env.js
+++ b/desktop/env.js
@@ -13,6 +13,8 @@ const dialog = require( 'electron' ).dialog;
  */
 const config = require( './lib/config' );
 const Settings = require( './lib/settings' );
+const EditorContextMenu = require( './lib/menu/editor-context-menu' );
+const GeneralContextMenu = require( './lib/menu/general-context-menu' );
 
 /**
  * Module variables
@@ -83,5 +85,8 @@ debug( '========================================================================
 // Define a global 'desktop' variable that can be used in browser windows to access config and settings
 global.desktop = {
 	config: config,
-	settings: Settings
+	settings: Settings,
+	editorContextMenu: EditorContextMenu,
+	generalContextMenu: GeneralContextMenu
 };
+
--- a/desktop/lib/menu/editor-context-menu.js
+++ b/desktop/lib/menu/editor-context-menu.js
@@ -34,11 +34,6 @@ var DEFAULT_MAIN_TPL = [{
   label: 'Paste',
   role: 'paste'
 }, {
-  label: 'Paste and Match Style',
-  click: function() {
-    BrowserWindow.getFocusedWindow().webContents.pasteAndMatchStyle();
-  }
-}, {
   label: 'Select All',
   role: 'selectall'
 }];
@@ -71,6 +66,7 @@ function getTemplate(val, defaultVal) {
     return cloneDeep(defaultVal);
   }
 }
+
 
 /**
  * Builds a context menu suitable for showing in a text editor.
--- a/desktop/lib/menu/editor-context-menu.js
+++ b/desktop/lib/menu/editor-context-menu.js
@@ -8,10 +8,10 @@
 
 var noop = function(){};
 var defaults = require('lodash.defaults');
-var isEmpty = require('lodash.isEmpty');
-var isFunction = require('lodash.isFunction');
-var isArray = require('lodash.isArray');
-var cloneDeep = require('lodash.cloneDeep');
+var isEmpty = require('lodash.isempty');
+var isFunction = require('lodash.isfunction');
+var isArray = require('lodash.isarray');
+var cloneDeep = require('lodash.clonedeep');
 var BrowserWindow = require('electron').BrowserWindow;
 var Menu = require('electron').Menu;
 
--- a/desktop/lib/menu/editor-context-menu.js
+++ b/desktop/lib/menu/editor-context-menu.js
@@ -9,8 +9,6 @@
 var noop = function(){};
 var defaults = require('lodash.defaults');
 var isEmpty = require('lodash.isempty');
-var isFunction = require('lodash.isfunction');
-var isArray = require('lodash.isarray');
 var cloneDeep = require('lodash.clonedeep');
 var BrowserWindow = require('electron').BrowserWindow;
 var Menu = require('electron').Menu;
@@ -48,27 +46,6 @@ var DEFAULT_SUGGESTIONS_TPL = [
 ];
 
 /**
- * if passed a function, invoke it and pass a clone of the default (for safe mutations)
- * if passed an array, use as is
- * otherwise, just return a clone of the default
- * @param val {*}
- * @param defaultVal {Array}
- * @returns {Array}
- */
-function getTemplate(val, defaultVal) {
-  if(isFunction(val)) {
-    return val(cloneDeep(defaultVal));
-  }
-  else if(isArray(val)) {
-    return val;
-  }
-  else {
-    return cloneDeep(defaultVal);
-  }
-}
-
-
-/**
  * Builds a context menu suitable for showing in a text editor.
  *
  * @param {Object=} selection - An object describing the current text selection.
@@ -84,15 +61,15 @@ function getTemplate(val, defaultVal) {
  *    Receives the default suggestions template as a parameter. Should return a template.
  * @return {Menu}
  */
-var buildEditorContextMenu = function(selection, mainTemplate, suggestionsTemplate) {
+var buildEditorContextMenu = function(selection) {
 
   selection = defaults({}, selection, {
     isMisspelled: false,
     spellingSuggestions: []
   });
 
-  var template = getTemplate(mainTemplate, DEFAULT_MAIN_TPL);
-  var suggestionsTpl = getTemplate(suggestionsTemplate, DEFAULT_SUGGESTIONS_TPL);
+  var template = cloneDeep(DEFAULT_MAIN_TPL);
+  var suggestionsTpl = cloneDeep(DEFAULT_SUGGESTIONS_TPL);
 
   if (selection.isMisspelled) {
     var suggestions = selection.spellingSuggestions;
--- a/desktop/lib/menu/editor-context-menu.js
+++ b/desktop/lib/menu/editor-context-menu.js
@@ -1,0 +1,122 @@
+
+/**
+ * Original library:
+ * Author: Jeff Wear, Mixmax, Inc
+ * License: MIT
+ * From: https://github.com/mixmaxhq/electron-editor-context-menu
+ */
+
+var noop = function(){};
+var defaults = require('lodash.defaults');
+var isEmpty = require('lodash.isEmpty');
+var isFunction = require('lodash.isFunction');
+var isArray = require('lodash.isArray');
+var cloneDeep = require('lodash.cloneDeep');
+var BrowserWindow = require('electron').BrowserWindow;
+var Menu = require('electron').Menu;
+
+
+var DEFAULT_MAIN_TPL = [{
+  label: 'Undo',
+  role: 'undo'
+}, {
+  label: 'Redo',
+  role: 'redo'
+}, {
+  type: 'separator'
+}, {
+  label: 'Cut',
+  role: 'cut'
+}, {
+  label: 'Copy',
+  role: 'copy'
+}, {
+  label: 'Paste',
+  role: 'paste'
+}, {
+  label: 'Paste and Match Style',
+  click: function() {
+    BrowserWindow.getFocusedWindow().webContents.pasteAndMatchStyle();
+  }
+}, {
+  label: 'Select All',
+  role: 'selectall'
+}];
+
+var DEFAULT_SUGGESTIONS_TPL = [
+  {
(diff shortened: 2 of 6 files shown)


mkaz:
Menu view, open to suggestions on what to include: 
right-click

jasmussen:
This is very cool!

I'm wondering — why limit this to only textareas? Can it be enabled for everything? Note: I'm not yet convinced enabling it for everything is a great idea, I'm just wondering at this point if it's technically possible. 

Also — this seems to work perfectly in `input` fields and `textarea`s. But it didn't actually work in the new editor, in visual mode. I'm not sure if that's due to some `contenteditable` or iframe `wizardry`, but otherwise, nice work!

package.json
```diff
@@ -48,6 +48,7 @@
   },
   "dependencies": {
     "debug": "^2.2.0",
+    "electron-editor-context-menu": "^1.1.0",
```
mjangda: Can we just pull the code into the desktop repo as a new module instead of adding as a dependency (especially since it's just one small file)? With proper credit + links, of course.

public_desktop/desktop-app.js
```diff
@@ -89,6 +89,24 @@ function startDesktopApp() {
 			ev.preventDefault();
 		}
 	}
+	
+	var remote = require( 'electron' ).remote;
+	var buildEditorContextMenu = remote.require('electron-editor-context-menu' );
+	function contextMenu( ev ) {
+		if ( ! ev.target.closest( 'iframe body, textarea, input, [contenteditable="true"]' ) ) {
```
mjangda: We use iframes in the reader so probably not great to target those.

public_desktop/desktop-app.js
```diff
@@ -89,6 +89,24 @@ function startDesktopApp() {
...
+			return;
+		} else {
+			console.log( "Target: " + ev.target.closest );
```
mjangda: Before we merge, should convert this to `debug`.

package.json
mkaz: Yeh, this was just to get the initial going - I think we'll want our own because will likely want a different context menu depending on what is clicked

public_desktop/desktop-app.js
mkaz: Still in testing, will get removed. It doesn't work anyways - context menu doesn't show in iframes

mkaz:
@jasmussen I think we can have a different context menu for non textareas - Cut / Copy / Paste doesn't make sense if you're not in a textarea, likewise for Undo/Redo. However, we could include Back/Forward or other useful menu items.

mjangda:
`Copy` is useful when you have selected text.

jasmussen:
> @jasmussen I think we can have a different context menu for non textareas - Cut / Copy / Paste doesn't make sense if you're not in a textarea, likewise for Undo/Redo. However, we could include Back/Forward or other useful menu items.

As Mo says, "Copy" is useful anywhere, even if you can't paste. Back/forward are nice. Slack also has these features, of which some feel useful:

screen shot 2016-03-09 at 09 42 26

mkaz:
spellcheck-test

jasmussen:
So good Marcus. So good.

mkaz:
The spellcheck fixes will be filed under issue #101

desktop/lib/menu/editor-context-menu.js
```diff
@@ -0,0 +1,118 @@
...
+ *    If it's a function, used to customize the template of always-present menu items.
+ *    Receives the default template as a parameter. Should return a template.
+ * @param {Function|Array} suggestionsTemplate - Optional. If it's an array, use as is.
+ *    If it's a function, used to customize the template of spelling suggestion items.
+ *    Receives the default suggestions template as a parameter. Should return a template.
+ * @return {Menu}
+ */
+var buildEditorContextMenu = function(selection, mainTemplate, suggestionsTemplate) {
+
+  selection = defaults({}, selection, {
```
mjangda: We can probably just use `Object.assign` here?

desktop/lib/menu/editor-context-menu.js
```diff
@@ -0,0 +1,118 @@
...
+ * License: MIT
+ * From: https://github.com/mixmaxhq/electron-editor-context-menu
+ */
+
+var noop = function(){};
+var defaults = require('lodash.defaults');
+var isEmpty = require('lodash.isempty');
+var isFunction = require('lodash.isfunction');
+var isArray = require('lodash.isarray');
+var cloneDeep = require('lodash.clonedeep');
```
mjangda: I wonder if we should drop these lodash libs except for `clonedeep` since there are easy-ish alternatives.

public_desktop/desktop-app.js
```diff
@@ -92,6 +94,42 @@ function startDesktopApp() {
 
 	debug = gGebug( 'desktop:browser' );
 
+	try {
+		var buildEditorContextMenu = electron.remote.require( '../desktop/lib/menu/editor-context-menu' );
+		var buildGeneralContextMenu = electron.remote.require( '../desktop/lib/menu/general-context-menu' );
```
mjangda: Might be nice to avoid the traversal and just be able to require `lib/menu/editor-context-menu`. We may need to set `NODE_PATH` for this to work?

public_desktop/desktop-app.js
```diff
@@ -92,6 +94,42 @@ function startDesktopApp() {
...
+	} catch (e) {
+		debug( "Error loading context menus", e.message);
+		loadContextMenu = false;
+	}
+	function contextMenu( ev ) {
```
mjangda: Can we move the functions up along with the others? They break the flow of the code which can be a bit confusing.

public_desktop/desktop-app.js
```diff
@@ -92,6 +94,42 @@ function startDesktopApp() {
...
+		// The 'contextmenu' event is emitted after 'selectionchange' has fired but possibly before the 
+		// visible selection has changed. Try to wait to show the menu until after that, otherwise the 
+		// visible selection will update after the menu dismisses and look weird. 
+		setTimeout(function() {
+			menu.popup(electron.remote.getCurrentWindow());
+		}, 30);
+	}
+
+	function resetSelection() {
+		selection = {
```
mjangda: FWIW, the name `selection` threw me off since it's different from the general context menu selection.

mjangda:
@mkaz what's the best way to test the spellchecker? It fails to initialize for me because of a version mismatch.

public_desktop/desktop-app.js
mkaz: Agree, also since it doesn't work in Windows. How do we set NODE_PATH within Electron?

mkaz:
@mjangda Where does it fail to initialize, Windows or Mac?
What version mismatch, with Node or Electron and Spellcheck?

For me it is working on Mac with no problems, does not work on Windows due to missing NODE_PATH issues

mjangda:
OSX. Looks like `electron.remote.require( 'spellchecker' )` fails with an exception: `Error: Module version mismatch. Expected 47, got 46.(…)`

mjangda:
Bugs:
- "Copy" is displayed and enabled even if I don't have any text selected. Clicking it does nothing.
- "Redo" is displayed even if there are no redo actions available. Presumably the same thing happens with Undo.
- Undo and Redo 
- Menu doesn't display when I right-click on the Visual editor.
- But: if I right-click on a misspelled word in the Visual editor and then right click on another input box (like the title), I get suggestions for the previous word from the editor.

mjangda:
Oh, and for posterity: need to use node 5.x to get the latest spellchecker bindings. Updating node versions also requires a `make distclean` before running.

mkaz:
Bugs from @mjangda 

- [ ] Copy displayed and enabled with none selected, OS X only. Disabled properly in Linux, unsure why it shows on Mac

- [x] - Undo/Redo menus removed

- [x] - Visual editor context menu fixed with https://github.com/Automattic/wp-calypso/pull/4063

- [ ] - Title field input not spell checking correctly, going to address in https://github.com/Automattic/wp-desktop/issues/119

desktop/lib/menu/editor-context-menu.js
```diff
@@ -0,0 +1,87 @@
...
+  role: 'cut'
+}, {
+  label: 'Copy',
+  role: 'copy'
+}, {
+  label: 'Paste',
+  role: 'paste'
+}, {
+  label: 'Select All',
+  role: 'selectall'
```
mjangda: Minor: we should fix the indents in this file (guessing the spaces are from the original lib).

desktop/lib/menu/editor-context-menu.js
```diff
@@ -0,0 +1,87 @@
...
+		} else {
+			contextMenu.unshift.apply(contextMenu, suggestions.map( function( suggestion ) {
+				return {
+					label: suggestion,
+					click: function() {
+						BrowserWindow.getFocusedWindow().webContents.replaceMisspelling(suggestion);
+					}
+				};
+			} ).concat( {
+			  type: 'separator'
```
mjangda: Minor: fix indent here.

public_desktop/desktop-app.js
```diff
@@ -90,8 +92,60 @@ function startDesktopApp() {
...
+	}
+
+
 	debug = gGebug( 'desktop:browser' );
 
+
+	debug( 'Setting up Context Menus' );
+	try {
+		var buildEditorContextMenu = electron.remote.require( '../desktop/lib/menu/editor-context-menu' );
+		var buildGeneralContextMenu = electron.remote.require( '../desktop/lib/menu/general-context-menu' );
```
mjangda: I think you mentioned that the traversal breaks on Windows? As a workaround, we could pass the menus through the `desktop` global like we do for settings (`desktop = electron.remote.getGlobal( 'desktop' ); ... desktop.settings.isDebug ()`).

mkaz: That is based upon this setting of desktop global variable, not related to a module
https://github.com/Automattic/wp-desktop/blob/master/desktop/env.js#L84

It could be a solution to load the context menus into a global and use, but should be able to use the `remote.require` method to do so. Just doesn't want to work on Windows 😕

mjangda: We could do something like this:

```
global.desktop = {
	config: config,
	settings: Settings,
	contextMenus: { // could also just be a wrapper lib that handles the requires
		editor: require( 'require( './lib/menu/editor-context-menu' ),
		general: require( 'require( './lib/menu/general-context-menu' ),
	}
};
```

and then with can just use `desktop.contextMenus.editor` instead of `buildEditorContextMenu`.

desktop/env.js
```diff
@@ -13,6 +13,8 @@ const dialog = require( 'electron' ).dialog;
  */
 const config = require( './lib/config' );
 const Settings = require( './lib/settings' );
+const EditorContextMenu = require( './lib/menu/editor-context-menu' );
+const GeneralContextMenu = require( './lib/menu/general-context-menu' );
```
mjangda: Minor: I think capitalization for modules is generally limited to modules with a constructor (or that use `new`).

desktop/env.js
```diff
@@ -83,5 +85,8 @@ debug( '========================================================================
 // Define a global 'desktop' variable that can be used in browser windows to access config and settings
 global.desktop = {
 	config: config,
-	settings: Settings
+	settings: Settings,
+	editorContextMenu: EditorContextMenu,
+	generalContextMenu: GeneralContextMenu
```
mjangda: Might be nice to have all the menus under one key (`{ contextMenus: { editor: {}, general: {} } }`) since we may add more in the future and this will help keep the global clean.

public_desktop/desktop-app.js
```diff
@@ -90,8 +92,60 @@ function startDesktopApp() {
 		}
 	}
 
+	function contextMenu( ev ) {
```
mjangda: Something for future: we could probably split this up into smaller functions.

- `handleContextMenu` or `contextMenuCallback` that determines which menu to show.
- `showEditorContextMenu` and `showGeneralContextMenu` which generate the menu and trigger it.
- `triggerContentMenu` which has the `setTimeout` call.

Once broken up, we can have the `mce-contextmenu` callback just trigger `showEditorContextMenu` (and would help avoid any issues with event objects coming from different contexts).

mjangda:
Added a few more suggestions; take 'em or leave 'em :)

Otherwise :ship: :ship: :ship:

desktop/env.js
mkaz: I tried but for some reason didn't work

public_desktop/desktop-app.js
mkaz: Good suggestions, created #121

desktop/env.js
mjangda: I think it's a race condition; we assign to the vars too early. 

This works: https://gist.github.com/mjangda/788bb4654dabbadb030d

PR #118, opened 2016-03-11
Preference to disable spell check

mjangda:
Add a new preference to allow disabling spellcheck.

Currently prompts for a restart since we can't un-spellcheck already spellchecked words so better to start with a clean slate.

Reasons outlined in #101

### DIFF
--- a/desktop-config/config-base.json
+++ b/desktop-config/config-base.json
@@ -59,6 +59,7 @@
 		"proxy-type": "",
 		"proxy-url": "socks5://localhost",
 		"proxy-port": "8080",
-		"proxy-pac": ""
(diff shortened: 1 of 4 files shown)


mkaz:
Looks good, I think the only thing will be to add disabling of the preference if not supported based on locale, so the user won't think the spellcheck should work. Depends on #101

mjangda:
Cool. Do we have a list of languages that the spellchecker lib supports?

mkaz:
It looks like it is system dependent and should support numerous dictionaries
Per: https://github.com/atom/node-spellchecker/issues/4

PR #124, opened 2016-03-18
Fix/119 spellcheck

mkaz:
Update spellcheck zips for each platform.
Use skip words to avoid contractions being marked as misspelling
Add build instructions instructions for creating spellcheck native

Note: `make distclean` before running to get the latest spellcheck binary.

### DIFF
--- a/public_desktop/desktop-app.js
+++ b/public_desktop/desktop-app.js
@@ -183,11 +183,12 @@ function startDesktopApp() {
 // hard code wrongly interpreted words in spellchecker
 // electron has a bug with parsing of contractions
 // Ref: https://github.com/atom/electron/issues/1005
-function skipWords( locale, text ) {
-	// locale == 'en-us' already skipped in setup
-	var words = [ 'ain', 'couldn', 'didn', 'doesn', 'hadn', 'hasn', 'mightn', 'mustn', 'needn', 'oughtn', 'shan', 'shouldn', 'wasn', 'weren', 'wouldn'];
-	if ( text in words ) { return true; }
-	return false;
+var wordsToSpellSkip = {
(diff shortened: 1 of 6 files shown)


public_desktop/desktop-app.js
```diff
@@ -180,6 +180,16 @@ function startDesktopApp() {
 	}
 }
 
+// hard code wrongly interpreted words in spellchecker
+// electron has a bug with parsing of contractions
+// Ref: https://github.com/atom/electron/issues/1005
+function skipWords( locale, text ) {
+	// locale == 'en-us' already skipped in setup
+	var words = [ 'ain', 'couldn', 'didn', 'doesn', 'hadn', 'hasn', 'mightn', 'mustn', 'needn', 'oughtn', 'shan', 'shouldn', 'wasn', 'weren', 'wouldn'];
```
mjangda: Probably good to future proof this by anticipating other languages:

```
words = {
    "en-us": [ 'ain', ... ],
}

return ( words[ locale ] && text in words[ locale ] );
```

Because we'll inevitably forget about this function when we remove the `en-us` restriction :)

Also, bonus, move the array to a global so we're not generating the array every time `skipWords` is called.

public_desktop/desktop-app.js
```diff
@@ -180,6 +180,16 @@ function startDesktopApp() {
 	}
 }
 
+// hard code wrongly interpreted words in spellchecker
+// electron has a bug with parsing of contractions
+// Ref: https://github.com/atom/electron/issues/1005
+function skipWords( locale, text ) {
```
mjangda: Better name: `shouldNotSpellcheckWord`

public_desktop/desktop-app.js
mjangda: Also, the `in` check doesn't work here. I think we need an `indexOf` instead: `return words.indexOf( text ) !== -1;` instead.

mjangda:
Looking good. Left a few comments.

public_desktop/desktop-app.js
mkaz: ah... right, its an array - which explains what I thought was erratic behavior

PR #125, opened 2016-03-18
Update exception handlers

mjangda:
Two changes:

1) Consolidate multiple exception handlers into one

An exception handler added early in development was more lenient with network errors, which can sometimes happen with updater. A handler added later overrode the existing one and failed on everything.

This combines the two so we handle network errors accordingly and properly fail on others.

We still run the exception handler super early to catch any issues during the bootup process.

2) Include system data in exception error message.

@mkaz can you take a look?

### DIFF
--- a/desktop/app-handlers/exceptions/index.js
+++ b/desktop/app-handlers/exceptions/index.js
@@ -11,6 +11,7 @@ const dialog = electron.dialog;
  * Internal dependencies
  */
 const crashTracker = require( 'lib/crash-tracker' );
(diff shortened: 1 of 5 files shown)


mkaz:
Looks good, though I wasn't able to trigger the pop-up. Each network scenario I tried, with proxy on/off, network or not all seemed to be handled with the yellow network banner alert or it worked as expected.

If you have a known test scenario to trigger I can try it, otherwise :+1:

PR #127, opened 2016-03-22
Update Calypso, and add Webpack loader for sections.js

ockham:
Fixes `TypeError: __webpack_require__(...).get is not a function`.

To get an updated, working app, we need:

* [x] ~~https://github.com/Automattic/wp-calypso/pull/4290~~
* [ ] https://github.com/Automattic/wp-calypso/pull/4148
* [ ] This PR (updated to the latest Calypso, including those two fixes)

/cc @mkaz @Tug

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ CALYPSO_JS := $(CALYPSO_DIR)/public/build.js
 CALYPSO_JS_STD := $(CALYPSO_DIR)/public/build-desktop.js
 CALYPSO_CHANGES_STD := `find "$(CALYPSO_DIR)" -newer "$(CALYPSO_JS_STD)" \( -name "*.js" -o -name "*.jsx" -o -name "*.json" -o -name "*.scss" \) -type f -print -quit | grep -v .min. | wc -l`
 CALYPSO_BRANCH = $(shell git --git-dir ./calypso/.git branch | sed -n -e 's/^\* \(.*\)/\1/p')
-WEBPACK_BIN := @$(NPM_BIN)/webpack
+WEBPACK_BIN := $(NPM_BIN)/webpack
 
 # sets to 1 if NPM version is >= 3
 NPMGTE3 := $(shell expr `npm -v | cut -f1 -d.` \>= 3)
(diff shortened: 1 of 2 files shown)


Tug:
Yeah I also tried bringing changes from `webpack.config.node.js` here and ran into the same error. Any idea how to resolve this?
Tell me if I can be of any help :)

ockham:
My hunch is that we're missing [this](https://github.com/Automattic/wp-calypso/blob/dff564cf9fe9c026995c82bd886717898a3f7db8/Makefile#L121), see https://github.com/Automattic/wp-calypso/tree/master/client/config#clientconfig

ockham:
Hmm, it's not that, since we're building that file through `package`'s dependency on `build-if-changed` -> `build-if-not-exists` -> build -> calypso `build`, before [invoking webpack](https://github.com/Automattic/wp-desktop/blob/master/Makefile#L85). So that file is present at that stage, but webpack can't seem to find it. Maybe it's some setting that is missing in `webpack.shared.js`, though I still fail to see what it could be.

Aside -- Having webpack run twice (once through Calypso's build process, and once for the desktop app) is somewhat confusing, and maybe the source of errors like this. IMO, we'd ideally run it only once.

mkaz:
I'm not sure if it helps, but I think I tracked it down to his commit, but I can't figure out why
https://github.com/Automattic/wp-calypso/commit/aa097b1f54efb2705e403459723659ca77d87dc5

If I do a `cd calypso; git checkout aa097b1f54efb2705e403459723659ca77d87dc5` and then build the desktop app I get the webpack error. Using the hash prior to that one in the commit log, which is `66e4bbff0e19befff22b9d82db73675bae32e018` I do not get the error.

With merging and commit dates things are quite as linear, but from what I can figure it looks like it started breaking on March 17

ockham:
I'm pretty sure it's because of Automattic/wp-calypso#4086, which does some Webpack magic to provide [this stuff](https://github.com/Automattic/wp-calypso/pull/4086/files#diff-edbb9ad814e9b0c3dac6e3a5a7f0d944R11).

I can get the desktop app to run with this PR, plus with [this line](https://github.com/Automattic/wp-calypso/blob/master/client/sections.js#L1) changed to `var config = require( './config' )` (note the `./`). However, the app is then only showing a blank page with the 'W' logo background -- but that might be a different issue, e.g. https://github.com/Automattic/wp-calypso/pull/4148

ockham:
> app is then only showing a blank page with the 'W' logo background

Same in `wp-calypso`, with `code-splitting` set to false. So I'm confident it's unrelated to this issue.

mkaz:
@ockham confirmed it works with that change - the style issue is different

ockham:
Had to revert #4290. Updated this PR to work without any Calypso modifications. Big props to @blowery for helping with that!

mkaz:
Confirmed, good fix. The path change probably is better long run too, since it'll behave more like Calypso which is picking it up in its path.

:ship:

ockham:
Tests were still broken, so I amended the last commit to also prepend `NODE_PATH=calypso/client` to the Webpack invocation in the `test` target.

mkaz:
@ockham Good catch! I was assuming it was still related to the blank issue on Calypso master

PR #135, opened 2016-03-28
README: add remote flag to submodule update cmd

jblz:
Without updating the wp-calypso dependency, all I get is a blank screen with a WordPress logo & this in the console:
```
Uncaught ReferenceError: preloadHub is not defined
desktop-app.js:140 Uncaught TypeError: window.AppBoot is not a function
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ WordPress.com for Desktop is an [Electron](https://github.com/atom/electron) wra
 1. Clone this repository locally
 1. Update the Calypso submodule with:
  - `git submodule init`
- - `git submodule update`
+ - `git submodule update --remote`
 1. Create a `calypso/config/secrets.json` file and fill it with [secrets](docs/secrets.md)
 1. `make run` to build and run the app
 


jasmussen:
Thanks for doing this!  👍

ockham:
The `--remote` flag is only needed to update the current Calypso submodule to the latest `master` as found in the `wp-calypso` repo, and record the corresponding subproject commit SHA. We currently need to do that in order to update the calypso dependency to a more recent version that doesn't have the previous bugs. However, we can simply commit that submodule SHA to `wp-desktop` in order to require that SHA for all `wp-desktop`.

People that check out or pull the `wp-desktop` repo can then update their `wp-calypso` submodule to that commit by simply running `git submodule update` (without the `--remote`).

tl;dr:

1. `git submodule update --remote`, and adding/committing to this repo to update the submodule to the latest `wp-calypso` commit
2. `git submodule update` (after a fresh clone or `git pull`) to update the local calypso submodule to that commit

So I'd suggest rather filing a PR with the updated `wp-calypso` dependency (1.), and maybe adding a note to 2. after `git pull`.

jblz:
Thanks, @ockham. That makes sense. I'll close this and start another PR w/ an updated SHA.

PR #136, opened 2016-03-28
Update wp-calypso submodule

jblz:
Update wp-calypso submodule to reference SHA1 [ea5faae702f0680a4976298bdb7a7169b3f4fae8](https://github.com/Automattic/wp-calypso/tree/ea5faae702f0680a4976298bdb7a7169b3f4fae8)

Without updating the wp-calypso dependency, all I get is a blank screen with a WordPress logo & this in the console:
```
Uncaught ReferenceError: preloadHub is not defined
desktop-app.js:140 Uncaught TypeError: window.AppBoot is not a function
```
See #135 -- `git submodule update --remote` was a workaround, but referencing a recent SHA1 checksum is a better solution.

I was able to run through a signup flow & clicked around a good amount without any obvious regressions.


mkaz:
Thanks, I'll usually update prior to a release - but since we're getting real close and this should include all Calypso changes for next release, we can base it off this sha

PR #144, opened 2016-03-30
bump wp-calypso version

rralian:
This PR just bumps to the latest wp-calypso version. This is necessary to pick up a fix in wpcom.js that was otherwise causing fatal errors in wp-desktop. I'll let @mkaz do all the tagging magic.


nylen:
See also:

See also:

- p1459293216000605-slack-wordpress-desktop
- https://github.com/Automattic/wpcom.js/pull/166
- https://github.com/Automattic/wpcom-unpublished/pull/10
- https://github.com/Automattic/wp-calypso/pull/4410

PR #146, opened 2016-03-30
Build: Add calypso's node modules as a cache directory

blowery:
This should speed up subsequent builds, as `npm install` for calypso will take less time


blowery:
Yup, this shaves about 2m:20s off builds that have a good cache. `make build` takes 2:25 vs 4:40 without cache.

mjangda:
Nice! :shipit:

PR #192, opened 2016-05-27
Ignore WP-Calypso's components usage stats

giuseppeg:
WP-Calypso master now builds a json file containing the ui components usage stats from `client/components`. However this file is not generated for WP-Desktop.

When building the desktop app @gwwar noticed the following error:

stats

What happened:
Webpack tried to pick up that file but since we don't build it for the desktop app it couldn't find it.

The fix is pretty straightforward:
I am adding the file `devdocs/components-usage-stats.json` to the webpack's `externas` object so that it doesn't get bundled.

### DIFF
--- a/webpack.shared.js
+++ b/webpack.shared.js
@@ -42,7 +42,8 @@ module.exports = {
 		// These are Calypso server modules we don't need, so let's not bundle them
 		'webpack.config',
 		'bundler/hot-reloader',
-		'devdocs/search-index'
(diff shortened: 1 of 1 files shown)


giuseppeg:
Not sure if it is necessary to add an `exclude` to the `json-loader` too.

gwwar:
👍 This looks like it clears up the stray error message. cc @mkaz

PR #202, opened 2016-07-06
Update classes used to target layout elements

mtias:
Calypso has updated classes related to the main layout components and rendered unnecessary some of the height CSS.


johngodley:
Joen is much more familiar with the reasons behind these CSS styles, but he won't be back for another month. I've checked the login screen and general Calypso usage and it all LGTM.

mtias:
Thanks. Gradually we'll remove all the custom CSS needed for the desktop app as Calypso itself inherits it.

PR #206, opened 2016-07-12
Set webpack directory context

johngodley:
On my machine `__dirname` is set to a full path which prevents the packaged builds from running.

On a machine where packaged builds work, `__dirname` is set to a path relative to the build directory.


johngodley:
Confirmed with @mkaz that it doesn't seem to affect his already working builds. Going to run with this and file it under the mystery box.

PR #209, opened 2016-07-27
Add instructions for adding oauth_client_id

lamosty:
I was just installing and running the desktop app, followed all the instructions and still got an error about missing `oauth_client_id`. The error/warning is coming from https://github.com/Automattic/wp-calypso/blob/master/client/auth/connect.jsx#L30 which was introduced in https://github.com/Automattic/wp-calypso/pull/4506.

After adding the `oauth_client_id`, the desktop app works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ WordPress.com for Desktop is an [Electron](https://github.com/atom/electron) wra
  - `git submodule init`
  - `git submodule update`
 1. Create a `calypso/config/secrets.json` file and fill it with [secrets](docs/secrets.md)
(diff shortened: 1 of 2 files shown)


lamosty:
Actually, it doesn't work as expected. I'm getting this error:

selection_089

Is it really necessary to set the `oauth_client_id` in `calypso/config/desktop.json`? :)

johngodley:
None of this should be necessary - something changed in Calypso that has affected the desktop app. I've put a small fix in place here:

https://github.com/Automattic/wp-calypso/pull/7121

Issue #219, opened 2016-08-04
Bug: Cannot find module '../screen'


MicBosi:
A user reported the following error while starting WP desktop app for Mac ("platform":"darwin","release":"15.6.0","version":"1.4.0","build":"release")

```
Error: Cannot find module '../screen'
at Module._resolveFilename (module.js:438:15)
at Function.Module._resolveFilename
(/Applications/WordPress.com.app/Contents/Resources/electron.asar/common/reset-search-paths.js:47:12)
at Function.Module._load (module.js:386:25)
at Module.require (module.js:466:17)
at require (internal/module.js:20:19)
at Object.defineProperties.screen.get [as screen]
(/Applications/WordPress.com.app/Contents/Resources/electron.asar/browser/api/exports/electron.js:83:14)
at setDimensions
(/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:30316:26)
at openWindow
(/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:30332:31)
at Object.module.exports.openAbout
(/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:30354:4)
at menuItems.splice.click
(/Applications/WordPress.com.app/Contents/Resources/app/desktop.js:148050:20)

System info:
{"platform":"darwin","release":"15.6.0","version":"1.4.0","build":"release"}
```

chrisfromthelc:
I've also had a user report the same error message.

sendhil:
I have a fix for this deployed in 1.5.0-beta2. Long story short something in Electron 1.1.1 is causing the require('electron').screen to fail and until I have more time to investigate I rolled back to 1.1.0.

PR #234, opened 2016-09-07
Update electron to latest (1.3.5)

sendhil:
I've updated electron to the latest version. Previously we were having issues were Electron post version 1.1.1 would cause [this](https://github.com/Automattic/wp-desktop/blob/e81c3d5eaaaa0270b3228e54a94bd74b35444255/desktop/lib/window-manager/index.js#L42) code to blow up with an error about being unable to find `../screen.js` (https://github.com/Automattic/wp-desktop/issues/219). Our temporary solution was to hold off on upgrading to post Electron 1.1.1 until we could figure out how to solve the issue. 

I haven't figured out 100% why the error was happening but I have a temporary workaround. For some reason in >= Electron 1.1.1 if we attempt to access `electron.screen` (or any other nested electron API) from outside of either the top level import or app.on('ready') the require blows up saying it's unable to find the corresponding file. For now pulling in the screen functions from within app.on('ready') seems to be a temporary workaround until I can get to the bottom of this. As a note, the reason I pulled screen in via app.on('ready') instead of a top level import is because in this particular case the screen API requires the app to be ready before it can be accessed. Other properties we'd need to access (e.g. `BrowserWindow`) can be imported at the top level like they currently are.

@mkaz - would you mind taking a look?


mkaz:
It now requires node 6.1.0 to build, which I think is probably related to Calypso updates.

I built and poked around checking: login, new post, reader, and stats and didn't see an issue on Mac or Windows. 👍

PR #245, opened 2016-10-27
Theme Showcase: Enable Server-Side Rendering

mattwondra:
This PR is a companion to [#8870 in wp-calypso](https://github.com/Automattic/wp-calypso/pull/8870/commits).

It syncs up the modules that needed to be mocked with NormalModuleReplacementPlugin to get SSR working on the Theme Showcase, so that the build doesn't break when the wp-calypso PR merges.


ehg:
Builds fine! LGTM, thanks!

PR #260, opened 2017-01-25
Webpack Config: Remove obsolete stub

ockham:
After updating to latest Calypso. See https://github.com/Automattic/wp-calypso/pull/10906 for some background.

To test: Make sure that the app still runs.


astralbodies:
I think #286 makes this PR obsolete. @ockham can you confirm?

ockham:
> I think #286 makes this PR obsolete. @ockham can you confirm?

Yes! Closing.

PR #262, opened 2017-01-29
Removing unnecessary check for length === 0

alejandronanez:


lancewillett:
Hi @alejandronanez Could you provide more details on why this change is needed?

PR #263, opened 2017-02-02
Build: Use the new minified names

blowery:
Needed to work with https://github.com/Automattic/wp-calypso/pull/11006


blowery:
cc @sendhil 

When we get the desktop to the point where Automattic/wp-calypso#11006 is merged, we'll need to update the Makefile with this PR.

PR #271, opened 2017-02-24
add flatpak support

LiliC:
This PR adds support for building [flatpak](http://flatpak.org/) for WordPress desktop, which should be runnable on any Linux distro.

This PR also introduces a `disable_updates` option, which is set by default to false, but due to Flatpaks nature we can't do updates that way. Let me know if you have other ideas how to handle this, will be happy to make adjustments to it. Thanks!

### Build and run:

For all requirements have a look at [electron-installer-flatpak.](https://github.com/endlessm/electron-installer-flatpak#requirements)

To build this locally call:

```
make package-flatpak
```
This gives you with a .flatpak file, which you can then install:

```
flatpak install --user --bundle release/flatpak/io.atom.electron.WordPressDesktop_master_x86_64.flatpak
flatpak run io.atom.electron.WordPressDesktop 
```

p.s. Your documentation is really great!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,10 @@ package-osx: osx
 package-linux: linux
 	@node $(THIS_DIR)/resource/build-scripts/package-linux.js
 
+package-flatpak: config-release-flatpak package
+	@node $(BUILDER) linux
+	@node $(THIS_DIR)/resource/build-scripts/package-flatpak.js
+
 distclean: clean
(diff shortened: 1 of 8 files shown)


astralbodies:
Sorry for the lack of activity on this PR. I'm going to close this for now as we're considering options on how we want to handle distribution going forward and that'll change several things.

PR #275, opened 2017-03-06
Framework: updated makefile to work with single build.js

samouri:
Along with the `wp-calypso` [PR here](https://github.com/Automattic/wp-calypso/pull/11785) (which has already landed), this PR seeks make `wp-desktop` compatible with the latest changes to wp-calypso (single build for non-development envs).

### Testing
1. Checkout ~~`try/framework/ssr-config-desktop-compat`~~ `master` Calypso branch.
2. Run `make test` to verify it works properly.
3. Run `make run` and check if it builds and works properly.

PR #276, opened 2017-03-09
Webpack: Add alias for "desktop"

spen:
Hi all!
I'm opening this up early and without actual implementation in the hope that it aids @samouri with #275.

The alias added in this PR allows paths like `lib/config` to be replaced with the more specific `desktop/lib/config`.
This will help avoid conflicting references between `desktop` and Calypso's own 'modules' - `client` and `server`.

#### Testing
Right now theres no usage of `desktop/xx` paths so to test the alias works, you'll have to change some manually before running `make run`.
A good example change would be to update `desktop/lib/app-instance`s internal deps to:
```js
/**
 * Internal dependencies
 */
const config = require( 'desktop/lib/config' );
const platform = require( 'desktop/lib/platform' );
```

#### notes
I plan to add aliases for `server` and `client` eventually but I'll open another PR for these once I've properly tested them.
I also plan on updating a bunch of references to use the `desktop` alias as this will help with my current project of merging `wp-desktop` in to `wp-calypso` - Either as part of this PR or in a separate one.

### DIFF
--- a/webpack.shared.js
+++ b/webpack.shared.js
@@ -47,6 +47,9 @@ module.exports = {
 		'devdocs/components-usage-stats.json'
 	],
 	resolve: {
+		alias: {
+			'desktop': path.join( __dirname, 'desktop' ),
+		},
 		extensions: [ '', '.js', '.jsx' ],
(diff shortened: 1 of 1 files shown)


samouri:
Yes! Thank you @spen, this is sweet.

I'll test this now

spen:
Closing this for now as I don't think this is quite the solution... I'm still tempted to add aliases but not until desktop is actually merged in to Calypso :)

Issue #237, opened 2016-09-09
NaN error code if the app opens on the My Sites tab


rachelmcr:
#### Steps to reproduce
1. Open the WordPress.com app
2. Open the My Sites tab
3. Quit the app
4. Reopen the app
#### Expected behavior

I expect the app to reopen on the same screen (My Sites) that I was on before.
#### Actual behavior

The app begins to load and I consistently get this error message:

> WordPress.com ran into an error! (code NaN)

![screen shot 2016-09-09 at 10 24 54 am](https://cloud.githubusercontent.com/assets/8658164/18382152/ade62430-7677-11e6-9c1b-7874febd4ca4.png)

(Note that the error screen has a broken image at the top.)

Using the "Try Again" button loads the app successfully, but I end up on the Reader tab.

**Bonus:** If I quit the app while on that error screen, and then reopen it, I get this error screen:

![screen shot 2016-09-09 at 10 17 58 am](https://cloud.githubusercontent.com/assets/8658164/18382051/471b76f6-7677-11e6-9b04-34df6c559974.png)

rachelmcr:
I also ran into the same error with these steps:

1. Opened the WordPress.com app
2. Opened the My Sites tab
3. Selected a self-hosted (Jetpack-connected) site
4. Noticed that the site has updates available (the yellow update icon appeared next to the site title in the sidebar)
5. Clicked the link to update, which opened a browser window with the site's WP Admin
6. In the app, the blue error screen (code NaN) appeared

Unfortunately, now I'm completely unable to open the My Sites tab in the app. Every time I open My Sites it defaults to that self-hosted site and I immediately get that blue error screen. Quitting and restarting the app doesn't not resolve it.

sendhil:
@rachelmcr - I'm trying to reproduce this - once you clicked on the My Sites tab did you click on any particular item?

sendhil:
Ahhh I got it to reproduce, n/m.

carinapilar:
Got an user today with the same problem, but on a different situation. He said when he tries to click on the Themes menu it shows this error. Ticket 2862317

tejas-shanthraj:
Running with the same error. When i try to customize my theme it gives me that error and when i click on try again takes me back to reader profile. Any idea how to get this working?

qwqclover:
the same error.

kevinvanryck:
The same thing is happening to me.

ehti:
This is still an issue. Another report in #232374-hc

[...]

PR #308, opened 2017-06-16
Preview: Ignore x-frame-options for iframes

marekhrabe:
This PR allows us to show preview even for sites that send `x-frame-options: sameorigin` or even `deny`. 

This header can be present on some Jetpack sites with plugins like this (https://wordpress.org/plugins/http-headers/) or the header can also be set in the code of some themes. Previously, loading of such site lead to the error with error code `NaN` and you had to restart the app in order to use Calypso again.

The header is not stripped on the server but rather on the network level inside of the Electron app thus we are not lowering the security of the site on the internet.

To test:
- have a jetpack site with plugin above mentioned
- in wp-admin, go to settings and enable `x-frame-options` to sameorigin
- test "Preview" functionality in the app

### DIFF
--- a/desktop/server/index.js
+++ b/desktop/server/index.js
@@ -60,6 +60,24 @@ function showAppWindow() {
 		}
 	} );
 
+	mainWindow.webContents.session.webRequest.onHeadersReceived( function( details, callback ) {
+		// always allow previews to be loaded in iframes
+		if ( details.resourceType === 'subFrame' ) {
+			const headers = Object.assign( {}, details.responseHeaders );
+			Object.keys( headers ).forEach( function ( name ) {
+				if ( name.toLowerCase() === 'x-frame-options' ) {
+					delete headers[ name ];
+				}
+			} );
+			callback( {
+				cancel: false,
+				responseHeaders: headers
+			} );
+			return;
+		}
+		callback( { cancel: false } );
+	} );
+
 	mainWindow.loadURL( appUrl );
 	//mainWindow.openDevTools();
 


astralbodies:
@marekhrabe I switched to this branch and tried following the steps to replicate and I'm not sure I'm seeing proper behavior. Before switching to this branch previewing a draft (from the post list) opened a web browser window with the `?p=` URL. Same thing happened from within the editor.

I switched to this branch and tried the same steps and got essentially the same behavior. In web Calypso I'm also seeing this behavior. Maybe my Jetpack 5.0 site isn't showing the same behavior with the http-headers plugin?

marekhrabe:
It seems we have a different starting point. My Jetpack site preview doesn't open externally but tries to load inside of the WebPreview modal and it leads to this screen:

preview

I don't really know why our websites behave differently. I know we have a flag `isPreviewable` on site object but I'm not sure what goes into determining that flag. Do you have https? My site is wpsandbox.me - yours?

marekhrabe:
I found this: https://github.com/Automattic/wp-calypso/blob/master/client/lib/site/computed-attributes.js#L49

marekhrabe:
I don't like the loop either. The problem was that the header key is not normalized and can show up as `x-frame-options` or `X-Frame-Options` or any other case.

spen:
I had a feeling that might be the case :)

marekhrabe:
@astralbodies do you want to have a final look before I merge? Last week Calypso Jetpack invites launched so I can add you to my site that you can easily use to replicate the problem :)

marekhrabe:
I noticed this might fix some issues from #237.

PR #314, opened 2017-07-12
Assert expectation of loc being string

yurynix:
When I started the application for the first time after installation, I've got that:

screen shot 2017-07-12 at 10 01 08

I'm not really sure how I landed there, since the stack seems to be from the `close` [event](https://github.com/Automattic/wp-desktop/blob/master/desktop/server/index.js#L87).


Tug:
I wasn't able to reproduce but the fix looks good!
🚢

PR #318, opened 2017-07-14
Use native calypso login instead of oauth2

Tug:
This PR updates wp-desktop to use the last version of calypso and switches the login method from oauth2 to the new WpLogin component (which relies on the few endpoints created at `/wp-login.php`). It relies on changes from https://github.com/Automattic/wp-calypso/pull/16187

It also updates the part which detects changes in calypso so you don't need to run `make distclean` anymore each time you changed something on calypso.

### Testing instructions
- Sandbox wordpress.com and apply patch D6385 and D6403
- Checkout this branch `git checkout update/login && git submodule update`
- Boot the app `npm start`
- Ensure that you can log in and log out properly
- Try to navigate to `/oauth-login` with `window.location = '/oauth-login';` to emulate a user who left at that page last time he loaded the app. Ensure you are redirected to `/log-in`

### Reviews 
- [x] Code
- [ ] Product

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ CERT_SPC := $(THIS_DIR)/resource/secrets/automattic-code.spc
 CERT_PVK := $(THIS_DIR)/resource/secrets/automattic-code.pvk
 CALYPSO_DIR := $(THIS_DIR)/calypso
 CALYPSO_JS := $(CALYPSO_DIR)/public/build.js
-CALYPSO_CHANGES_STD := `find "$(CALYPSO_DIR)" -newer "$(CALYPSO_JS)" \( -name "*.js" -o -name "*.jsx" -o -name "*.json" -o -name "*.scss" \) -type f -print -quit | grep -v .m. | wc -l`
+CALYPSO_CHANGES_STD := `find "$(CALYPSO_DIR)" -newer "$(CALYPSO_JS)" \( -name "*.js" -o -name "*.jsx" -o -name "*.json" -o -name "*.scss" \) -type f -print -quit | grep -v .min. | wc -l`
 CALYPSO_BRANCH = $(shell git --git-dir ./calypso/.git branch | sed -n -e 's/^\* \(.*\)/\1/p')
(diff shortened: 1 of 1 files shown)


yurynix:
First boot got:
```
> document.location.pathname
"/Users/yury/Automattic/wp-desktop/public_desktop/failed-to-start.html"
```

screen shot 2017-07-16 at 15 41 10

In the case when there is no network in the following [code](https://github.com/Automattic/wp-calypso/blob/a70587157b33783ba4d5e0e4a1e136680d459d51/client/boot/app.js#L49):
```
window.AppBoot = function () {
	var user = (0, _user2.default)();
	if (user.initialized) {
		boot(user);
	} else {
		user.once('change', function () {
			return boot(user);
		});
	}
};
```
will get stuck with calypso in "empty" state, for example, try not to apply D6385:
screen shot 2017-07-16 at 16 02 13


I don't think it's a regression of this PR, otherwise I were able to log-in and out just fine 👍

Tug:
Yep, it's not a regression, and I don't find it too problematic actually, we have this helpful message displayed :)

> First boot got:
> -> document.location.pathname
> "/Users/yury/Automattic/wp-desktop/public_desktop/failed-to-start.html"

We should investigate this, I think I've seen it before.

Tug:
Couldn't reproduce it. I cleaned the app cache but wasn't able to see that first screen so it might have been some problem in your build.

stephanethomas:
Shouldn't we bump the version number (see https://github.com/Automattic/wp-desktop/pull/311 as an example)?

Tug:
Right 👍 
I'll do that before merge, I'll need to update the submodule reference once we merge https://github.com/Automattic/wp-calypso/pull/16187 anyway!

stephanethomas:
The code looks good.

pento:
This needs to wait for Automattic/wp-calypso#16738 before the submodule can be updated, and the PR merged.

astralbodies:
@Tug Switched base to `develop` from `master`.

PR #322, opened 2017-07-31
Electron: Use 1.6.11

spen:
### * Blocked *
By https://github.com/Automattic/wp-desktop/pull/323 & https://github.com/Automattic/wp-desktop/pull/325, both are very minor changes that are needed to be able to run the make processes necessary for testing this PR on a new set up. You might not see the makensis issue if you've built previously and already have one of the checked-for versions installed.

Also, I'm unable to test the linux build at the moment, it builds successfully and I would assume it runs as the other 2 platforms do but I'd ❤️  if somebody else could test and confirm that for me 😃 

### Summary

This PR brings the version of Electron used in the project up to `v1.6.11`.

The motivation for doing so comes from the need to use node `v6.9.0` or above to remove a potential vulnerability. Ping me privately if you're not already aware of the details.
Electron `v1.6.11` uses node version `7.4.0`, so this issue will be solved.

### Test Plan
I have yet to test all that I'd like to as I'm having issues with certain builds. I'm on a relatively new machine and I think the issues that I'm seeing are more likely to be down to the machine rather the updated Electron version.
I'll keep working on successfully creating builds but I'd appreciate it if somebody else can beat me to it :)

- [x] Ensure App is built to use node `v7.4.0` in the main process
    - An easy way to check is to `console.log( process.versions[ 'node' ] )` in `desktop/env.js` or similar.

- [x] Successfully run `make run`
- [x] Successfully build and test `make package`
- [x] Successfully build and test `make package-linux`
    - [x] build
    - [x] test
- [x] Successfully build and test `make package-osx`
    - [x] build
    - [x] test
- [x] Successfully build and test `make package-win32`
    - [x] build
    - [x] test

### Notes

I was expecting to need to update other Electron related dependancies (`electron-builder`, `electron-packager` etc.) but so far that doesn't seem to be the case.
I'll aim to do some research around these dependancies to make sure they are not using an older node version themselves, or to see if there are other benefits to updating them all at once.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "electron-builder": "2.8.4",
     "electron-mocha": "^2.1.0",
     "electron-packager": "^7.0.2",
-    "electron": "1.3.5",
+    "electron": "1.6.11",
     "electron-rebuild": "^1.2.1",
     "eslint": "^2.10.2",
     "eslint-plugin-react": "^5.1.1",


spen:
Currently it seems that CircleCI tests fail due to an unrelated issue with the `.min.js` file extension during the `package` task.
PRs #323 and #324 both aim to fix that failing test (with different approaches). Once that test is fixed I'll mark as ready for review :)

astralbodies:

spen:
Before going ahead with updating Electron I'm going to explore other versions that include a node version that both includes the necessary fix and is on [nodes Long Term Support](https://github.com/nodejs/LTS).
If I fail to find one I believe we may be able to find one that is close enough and set the node version ourselves - though this could cause more trouble than it solves...

spen:
Update: All 3 platform packages are being built successfully - I just need to organise a way to test them :)

spen:
#323 now merged, If you're on a version of makensis that is already included in the makefile check - this is ready to test :)

astralbodies:
@spen Can you get current on what's in `develop` right now? I'm trying to make sure the build passes on CircleCI before merging.

astralbodies:
Successful on Ubuntu 17.0.4 - merging this! Thanks @spen

aidvu:
@spen quick question, any worries here about using different `node/npm` than Calypso which is then built and bundled with `7.4.0/4.0.5`?

Didn't really try to build it with `6.11.2`, but can do and see how it is? Thoughts?

/cc @astralbodies

spen:
Sorry @aidvu, I'm not sure I'm following properly so let me know if I'm mistaken here:

Do you mean are there worries with using Electron v1.6.11 (which packs v7.4.0 of node) and a different version of node to build Calypso?

If so, this is completely fine, the version that Electron uses (v7.4.0 in this case) is what ships out with the app itself (since we can't bet on the user having node installed or installed with the right version). The version we use in Calypso, or even in wp-desktop as our engine is (as you'll know) just to build and run. So the node that is referenced by Electron and the one referenced in each `package.json` are sort of inconsequential to each.

aidvu:
@spen what I mean is, we build both Calypso and the Electron app. And we have to use 7.4.0 for building Electron, which means Calypso gets built with the same version of it, and npm 4.0.5. :) Makes sense?

spen:
Got ya!
Honestly, this is something that I hadn't thought much about before.
I had always assumed that each project defined it's own NPM version so we're safe but thats not true. When running `make run` or `make build` from `wp-desktop` we `cd` into `wp-calypso` and run it's build script. Problem is that we're on `wp-desktop`s node version when we do this, so `wp-calypso` is built with a version of node that could be the same as is defined in it's `package.json` but may not be (As it stands, it's not the same - calypso: `6.11.2`, desktop: `6.10.3`).
For the most part, we're likely to avoid seeing any real issues but there is always the risk that what works in wp-calypso doesn't work in wp-desktop because of inconsistencies between the two versions of node referenced in either.

I think the ideal solution would be to have parity of `node` & `npm` versions between both projects, consistently.

It could even be worth performing a check that those versions match before allowing a build, at least for a release.

Sorry if the above seems confusing or unclear - having a tough time explaining! Feel free to ask for a better explanation!

aidvu:
@spen sounds good. If the bug you originally fixed is also fixed in an early Electron version, we can revert back to `6.11.2.`. :)

I'm up for testing later today to have these match. :+1:

spen:
The bug fixed here is independent of any node/npm version referenced in the `package.json` of either project so feel free to edit those :) So long as Electron stays on v1.6.11 or above we should be good!

aidvu:
Based on https://github.com/electron/electron/blob/v1.6.11/.node-version, I don't think it will be possible. What a weird decision to not go with LTS, but the current version.

I guess we'll be looking for the same versions after the next LTS release, bah... I recall now that I tried Calypso `node/npm` yesterday, and got `electron` failed to build.

spen:
Ahh but the version referenced there is the version that's packed up in the app. So you can build (as in running a build process) an Electron project with, say, node v6.x and it'll still pack up node v7.4.0 as part of the app.

aidvu:
Hmm, interesting. Will try again today. Might work. :)

spen:
I believe that to avoid any discrepancies between building Calypso from `wp-calypso` vs building it from `wp-desktop`, the versions should match those found in [Calypso's package.json](https://github.com/Automattic/wp-calypso/blob/master/package.json#L167-L170).

```json
"engines": {
  "node": "6.11.2",
  "npm": "3.10.10"
},
```

That way we can be confident that `make build` will be running `cd calypso && npm run build` with the versions of node and npm that calypso expects to be built with.

I'm not 100% confident that I'm correct here though (maybe hovering at 90%) so I'd love to get input from others first - @marekhrabe, do you have any insight here?

spen:
Just adding a notes here:
[This issue](https://github.com/electron/electron/issues/3068) describes how to get the various versions in Electron.
It'd be useful to log these from the main process (think 'server-side') in the app and keep an eye out for the output in the terminal.
What you should see is that this prints something like `v7.4.0` regardless of the node version used to run the build processes - it'd be worth trying changing `wp-desktop`s versions to those mentioned above (`"node": "6.11.2",  "npm": "3.10.10"`) and watching the output.

PR #323, opened 2017-07-31
Makefile: use .m.js file target in package task

spen:
While working on #322 I noticed that running `make package` or it's related tasks would throw an error while trying to copy a minified file. The file is there but with a `.m.js` extension rather than `.min.js`.
This PR fixes that :)

To test, run `make package` and ensure there are no errors thrown during the process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ package: build-if-changed
 	@cp $(CALYPSO_DIR)/config/_shared.json $(BUILD_DIR)/calypso/config/
 	@cp $(CALYPSO_DIR)/config/desktop.json $(BUILD_DIR)/calypso/config/
 	@rm $(BUILD_DIR)/calypso/public/style-debug.css*
-	@mv $(BUILD_DIR)/calypso/public/build.min.js $(BUILD_DIR)/calypso/public/build.js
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -107,7 +107,7 @@ package: build-if-changed
 	@cp $(CALYPSO_DIR)/config/_shared.json $(BUILD_DIR)/calypso/config/
 	@cp $(CALYPSO_DIR)/config/desktop.json $(BUILD_DIR)/calypso/config/
 	@rm $(BUILD_DIR)/calypso/public/style-debug.css*
-	@mv $(BUILD_DIR)/calypso/public/build.min.js $(BUILD_DIR)/calypso/public/build.js
+	@mv $(BUILD_DIR)/calypso/public/build.m.js $(BUILD_DIR)/calypso/public/build.js
```
jblz: We'll have to change this back the next time we update the wp-calypso submodule, correct? Maybe we should just try that instead and see if nothing breaks :)

jblz: Let's give it a go -- see how #324 treats ya

astralbodies:

spen:
Thanks for the heads up @astralbodies :)

PR #328, opened 2017-08-09
Use Calypso's npm scripts instead of make targets

ockham:
Companion PR to https://github.com/Automattic/wp-calypso/pull/17033.

(Doesn't get rid of `wp-desktop`'s `Makefile`, just updates it to use Calypso's `npm` scripts.)

To test:
* `cd calypso; git checkout master; git pull; cd..`
* `make run`


astralbodies:
@ockham Can you get current from develop? CircleCI failed and I think it's because of the build.m.js change.

astralbodies:
I ended up rebasing against `develop` - will merge once CircleCI is green.

PR #330, opened 2017-08-14
Fixes build.min.js kerfluffle and gets current on Calypso

astralbodies:
`build.min.js` is the most current and correct build file post-Webpack 3. This resolves the confusion that I may have caused in my PR merges and also gets current on wp-calypso.


astralbodies:
@spen Ignore the Travis CI integration failures - I apparently enabled that integration a tad on the early side and never adding the proper configuration for it to work right.

astralbodies:
Thanks for the review @spen and helping me get this resolved!

PR #338, opened 2017-10-12
Build: Compare node & npm versions between wp-desktop & Calypso

spen:
### Summary

This PR aims to add an extra check to the desktop build process in order to avoid building Calypso with versions of `node` or `npm` that differ from those which Calypso expects to be built with as doing so could cause issues.

### Current output:

screen shot 2017-10-14 at 22 13 08

### Testing
Make sure you have Calypso set up in the`/calypso` directory (follow `README.md` for this).

- Start by changing the `node` and `npm` versions under `engines` in `wp-desktop`s `package.json`...
- Try setting both to be different from those found in Calypso's `package.json`.
- run `make build` to attempt to start a desktop build.
- The process should bail almost immediately prompting you to update `wp-desktop`s versions - You should see that both `node` and `npm` versions appear with an `x` next to them.

Repeat the above process, but testing cases where 
- only one of the engines are wrong - you should see a green tick next to the matching engine.
- both `node` and `npm` are matching between projects - you should see the build process continue as usual, without errors thrown and without the above messages shown.

### Extra Thoughts

At this point, all we're checking for is that the expected NPM and Node versions are both equal between the two projects. We could further this by checking the developers current working versions by looking at the output of `node -v` & `npm -v`.
We might even consider being strict about one type of check (let's call the two checks 'working version' and 'project parity'). 
For instance, I personally have both `npm` and `node` running at the required versions so Calypso would build as expected even though the 'project parity' check would fail - should I be blocked from building until `wp-desktop`s `package.json` is updated? In that case a warning might suffice.
If, conversely, I've updated `wp-desktop`s versions to pass the 'project parity' check but have not upgraded my working versions this would still pose the risk of building Calypso with an unexpected version of `node`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ NPM ?= $(NODE) $(shell which npm)
 NPM_BIN = $(shell npm bin)
 
 RED=`tput setaf 1`
+GREEN=`tput setaf 2`
 RESET=`tput sgr0`
 
 START_APP := @$(NPM_BIN)/electron .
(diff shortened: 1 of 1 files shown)


spen:
Notes on [check-node-version](https://github.com/parshap/check-node-version/):
In Calypso, we use [check-node-version with --package](https://github.com/parshap/check-node-version/blob/master/usage.txt) to compare the working versions against those in the `package.json`. We couldn't use this directly from `wp-desktop` as it stands since that would be checking it's own `package.json` which doesn't necessarily represent Calypso's expected versions.
We may, however, be able to call a script from Calypso itself to do this check and use it's output.

spen:
screen shot 2017-10-15 at 12 33 23

I feel that it would make sense to throw an error in the above case, where the working versions don't match Calypso's versions... Would it still make sense to check for `wp-desktop` - `wp-calypso` parity and just give a warning for any mismatch there?

astralbodies:

astralbodies:
Thanks for the contribution @spen - I thought I had merged this and found a version discrepancy between wp-desktop and wp-calypso during this code freeze. Merging!

spen:
Thanks a heap @astralbodies :) Let me know how it pans out and I'll happily work on some improvements if necessary!

Issue #339, opened 2017-10-13
v2.9.0 build, WordPress.com ran into an error "Cannot find module '../screen'" and "TypeError"


ehti:
User reported in #731031-zen that they're getting the following error when opening the app:

<img src="https://user-images.githubusercontent.com/6505293/31525292-ab456e70-afd8-11e7-94d5-135b09f86c7f.jpg" height="400">

I was able to replicate it on Windows 10, latest creators update and latest WP app build (2.9.0).

To replicate:
1. Open the app, it might open fine
2. Click the close button, it'll go to tray
3. Double-click the app icon on desktop (not on tray), the error should show up

Also, I ran into the "TypeError" below when I was trying to close + quit (from tray) and then relaunching the app a few times. This error showed up after about 4-5 such quits/re-opens:

<img src="https://user-images.githubusercontent.com/6505293/31525179-d11e49b0-afd7-11e7-9ce3-ba91223ebe49.jpg" width="500">

rachelmcr:
Noting @Tug's findings here:

> Seems relevant: https://github.com/electron/electron/issues/8862

> So I guess `app.exit()` instead here https://github.com/Automattic/wp-desktop/blob/develop/desktop/lib/app-instance/index.js#L34

gemmagarner:
Another report in #740787-zen

gemmagarner:
#879939-f (a fresh install of 2.9.0 fixed the issue).

gemmagarner:
Reported in #740984-zen

Tug:
I could not reproduce the issue using the `master` branch on macos. I tried launching the app twice but didn't get any error (the second process stopped and made the first window appear again.
So it seems to be limited to windows...

GeoJunkie:
Reported in 739615-zen . User tried uninstall/reinstall twice without success.

gemmagarner:
Reported in 743022-zen (they can use the app after closing the message).

Lochlaer:
Another user is reporting a similar/not quite _exactly_ the same error in 753355-zen:

![wordperss error](https://user-images.githubusercontent.com/6569475/31948533-cb153402-b89c-11e7-9758-7a714b59bda2.JPG)

I've asked them to uninstall and then reinstall the app, and am also asking for specific details of their PC.

Lochlaer:
One more report here in 754366-zen. Error message:

[...]

PR #344, opened 2017-10-18
Fix showing error when user is trying to start a new instance of the app

Tug:
Fixes https://github.com/Automattic/wp-desktop/issues/339

### Testing Instructions
- Apply this patch
- `make clean`
- `make run`
- Once the app has booted, type `node_modules/.bin/electron .` in a new terminal
- Ensure no error is displayed in the terminal

### Reviews
- [ ] Code
- [ ] Product


astralbodies:
Switched the base on this PR to hit release/3.0.0 which is where I think you branched off of. 😄

PR #354, opened 2017-11-07
Fix external URL parsing

johngodley:
Use node’s URL parsing when checking external links

Replace regexes with Node’s URL parsing. This ensures we properly parse the URLs and miss out on any regex based fun.


astralbodies:

PR #355, opened 2017-11-07
Prevent links being dropped onto window

johngodley:
Links are prevented, but images should go through.

Note that Calypso has it's own handling on the media page and already blocks links.


astralbodies:

PR #356, opened 2017-11-09
Fix error icons and preload

johngodley:
Fix missing error icons

Move app setup into preload


spen:
It seems that the `boot` test is failing - investigating this now :)

spen:
I've found it difficult to track down the exact reason but it seems that either the event or listener that sends `is-calypso` or `is-calypso-response` is not firing/listening.

I remember having issues tracing the sender of `is-calypso` in my previous tasks - @johngodley do you know where this comes from?
The tests & listeners were introduced in #43 if that helps.

johngodley:
Ah, I forgot about the tests. I can't remember where anything comes from so I'll look into this some more!

johngodley:
@spen the new `preload.js` wasn't being included in the built file so it wasn't starting Calypso - unit test FTW! I've moved it into the public_desktop directory and it works locally for me

astralbodies:
There's a merge conflict @johngodley - are you able to check it out and see how to resolve it?

astralbodies:
Thanks!!

PR #363, opened 2017-11-13
Try: Dynamically use calypso's node version on CI

sirreal:
Dynamically detect the desired Node (nvm) version based on what's defined in `calypso/.nvmrc`.

This appears to work correctly based on [CI results for this branch](https://circleci.com/gh/Automattic/wp-desktop/11048):

---

```sh
$ nvm install $(cat calypso/.nvmrc)
# Exit code: 0
nvm install $(cat calypso/.nvmrc)
Downloading and installing node v6.11.2...
Downloading https://nodejs.org/dist/v6.11.2/node-v6.11.2-darwin-x64.tar.gz...

# …

$ nvm use $(cat calypso/.nvmrc)
# Exit code: 0
nvm use $(cat calypso/.nvmrc)
Now using node v6.11.2 (npm v3.10.10)
```

Issue #284, opened 2017-04-04
WordPress.com ran into an error


boostmoth:
A user forwarded us this crash report. No other information. See ticket 3141595

![image001](https://cloud.githubusercontent.com/assets/26887976/24638048/748a15b4-189a-11e7-9bea-64111439229f.jpg)

spen:
Seems to relate to #248

designsimply:
Thank you for the report! 

Closing this one as a duplicate of #248.

PR #364, opened 2017-11-13
Update to node 8

sirreal:
Node 8 is now LTS. Calypso upgrade is in progress.

Update wp-calypso accordingly.

Related: https://github.com/Automattic/wp-calypso/pull/19317 and #363


astralbodies:
You may want to rebase or merge in the latest from develop as I just merged in the latest changes from the release/3.1.0 branch.

sirreal:
👌  Thanks, rebased.

.nvmrc
```diff
@@ -1 +1 @@
-v6.11.5
```
sirreal: What do you think about symlinking this file to point to `calypso/.nvmrc`?

astralbodies: That makes sense! Is that supported on Windows?

sirreal: > That makes sense! Is that supported on Windows?

I don't think so 🙁 

But…

> Note: nvm does not support Windows (see #284). Two alternatives exist, which are neither supported nor developed by us […]

☝️  https://github.com/creationix/nvm#important-notes

So maybe we don't care about Windows support for `nvm` related stuff?

astralbodies: > So maybe we don't care about Windows support for nvm related stuff?

Yup that sounds accurate then if nvm isn't supported. 😄

astralbodies:
I added a commit to move Calypso ahead to latest master so the Node 8.9.1 changes were pulled in. I think the unit test will still fail for this PR because the CircleCI changes are in a different PR. I'm running the tests locally now and will merge when they pass.

sirreal:
> I think the unit test will still fail for this PR because the CircleCI changes are in a different PR

Maybe cherry-pick addb79d731b3896c6711acb2779c993fe88060ec and close #363

PR #379, opened 2017-12-06
Switch back to Linux for builds

astralbodies:
This is a temporary solution to get canary builds working for `wp-calypso`.


astralbodies:
I believe `npm test` is still executed on PRs in `wp-desktop` - https://circleci.com/gh/Automattic/wp-desktop/11121

It does look like the hook back into `wp-calypso` is called every time (even for PRs from wp-desktop` so we'll need to address that separately.

hoverduck:
Ah, thanks.  I forgot that it does that by default if you don't have an `override` step in the config, and I missed it when skimming the CI output.

PR #406, opened 2018-01-22
Fix: remove add-module-exports

samouri:
See: https://github.com/Automattic/wp-calypso/pull/16057

--

Is there a better way?

### DIFF
--- a/webpack.shared.js
+++ b/webpack.shared.js
@@ -3,6 +3,14 @@
  */
 var path = require( 'path' );
 var webpack = require( 'webpack' );
+var fs = require('fs');
+
+// disable add-module-exports from the sub-repo calypso's babelrc by modifying the file
+// @TODO ASAP: remove these lines when add-module-exports is removed from the babelrc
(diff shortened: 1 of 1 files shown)


astralbodies:
I'm definitely not the one to do a technical code review on this one - it compiles and runs at least. 😄

webpack.shared.js
```diff
@@ -3,6 +3,14 @@
  */
 var path = require( 'path' );
 var webpack = require( 'webpack' );
+var fs = require('fs');
+
+// disable add-module-exports from the sub-repo calypso's babelrc by modifying the file
+// @TODO ASAP: remove these lines when add-module-exports is removed from the babelrc
+const pathToBabelConfig = path.join( __dirname, 'calypso', '.babelrc' );
+const babelConfig = JSON.parse( fs.readFileSync( pathToBabelConfig, {encoding: 'utf8'} ) );
```
dmsnell: why not just do `const babelConfig = require( pathToBabelConfig )`

PR #417, opened 2018-02-09
Fix sharing page in new window

johngodley:
This allows the sharing popups to open in a new desktop window, instead of pushing them to a browser window. The key change is that the `DONT_OPEN_IN_BROWSER` needs to work for any URL on `http://127.0.0.1:41050/`, not just the root. This then allows the popup sharing connection windows to open.

On top of this we need to add the preload script otherwise the newly opened window doesn't have access to Electron, and will crash.

Additionally I added an extra check that the ports match when checking URLs. I don't think this is necessary, but it seems like a good thing to do.

To test:
- Open the post editor and click the 'add new service' in the sharing panel. A new window inside the desktop app should open and show all the sharing buttons

Note that after doing this PR and thinking further, I don't know if this is the best thing to be doing as you've now got two desktop windows open, and can go and do other things in the newly opened window. This is probably something that needs changing in Calypso to either not show this on desktop or to not open in a new window on desktop. It's a very browser-based action anyway and not really suitable for the desktop app.

### DIFF
--- a/desktop/window-handlers/external-links/index.js
+++ b/desktop/window-handlers/external-links/index.js
@@ -6,6 +6,7 @@
 const shell = require( 'electron' ).shell;
 const debug = require( 'debug' )( 'desktop:external-links' );
 const { URL } = require( 'url' );
+const path = require( 'path' );
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


nsakaimbo:
Stale - closing.

PR #421, opened 2018-03-02
[WIP] Restructure build process

adlk:
We are moving away from the makefile setup and moving to npm scripts to harmonise wp-desktop and wp-calypso. It should become easier to build and deploy new releases and developers should be able to easily test their changes directly in the app without cumbersome manual rebuilds.

**Note:** This PR is not yet integrated in any way in [wp-calypso](https://github.com/Automattic/wp-calypso). 

**Note:** At the moment there is an ongoing evaluation if calypso will move back to makefile.

## Prerequisites
1. Copy `calypso/config/desktop.json` to `calypso/config/desktop-development.json`
2. In your new config file, change `"env": "production"` to  `"env": "development"` (Not the complete solution, see bugs)

## How to run wp-desktop
```bash
$ NODE_ENV=local CALYPSO_ENV=desktop npm run start
```

## How to run wp-desktop in development mode
Don’t forget to run `$ nvm use` in **every** terminal session.
```bash
# Terminal 1 - Compiling main process scripts
$ NODE_ENV=development npm run build:desktop

# Terminal 2 - Running calypso in dev mode
$ NODE_ENV=development CALYPSO_ENV=desktop-development npm run start:calypso

# Terminal 3 - Start electron
$ NODE_ENV=development npm run start:electron
```

I am aware that it is cumbersome to have 3 active terminal sessions to do this. The main reason that is preventing me from using `concurrently` is that I'm experiencing issues with inheriting env vars in the npm scripts.

## How to build wp-desktop
- Work in Progress 👷‍♀️👷‍♂️
- More information: #425 

## Todos:
- [x] Enable dev mode
- [ ] Find a solution to properly set a default env and allow npm scripts to inherit env vars. Can we do this easily with npm scripts?  
- [ ] Find a proper way to run `dev` scripts concurrently to avoid 3 terminal sessions
- [x] Move to new electron-builder implementation
- [ ] Remove unused build helper scripts
- [ ] Clean up npm scripts & dependencies
- [x] (Re-) Configure CircleCI or Travis + Appveyor
- [ ] _(to be extended)_

## Bugs:
- [x] ~~Calypso: requests  to resources like css and javascript files take around 20 seconds to resolve + hot reloader times out~~ Fixed: This was a simple misconfiguration in the `desktop-development` calypso config. Will share a `desktop-development.json` once the todo _"Update calypso `desktop` config..."_ is done.


resource/build-scripts/build-config-file.js
```diff
@@ -1,15 +1,29 @@
-'use strict';
-/**
- * External Dependencies
- */
-var base = require( '../../desktop-config/config-base.json' );
-var config = require( process.argv[2] );
+// TODO: evaluate if there is a more straightforward way to create config files e.g. how does calypso solve this?
```
ockham: Yes please! https://github.com/Automattic/wp-calypso/tree/master/server/config and https://github.com/Automattic/wp-calypso/blob/master/client/lib/create-config/index.js seem relevant 🙂

resource/build-scripts/build-config-file.js
```diff
@@ -1,15 +1,29 @@
...
+const CONFIG_SRC_DIR = '../../desktop-config/';
+const CONFIG_DEST = './desktop/config.json';
+
+const path = require( 'path' );
+const fs = require( 'fs' );
+const chalk = require( 'chalk' );
+
+const baseConfig = require( path.join( CONFIG_SRC_DIR, 'config-base.json' ) );
```
dmsnell: it's kind of silly to use `path.join()` with this `CONFIG_SRC_DIR` constant when the constant is defined without `path.join()`

```js
const CONFIG_SRC_DIR = path.join( __dirname, '..', '..', 'desktop-config' );
const CONFIG_DEST = path.join( __dirname, 'desktop', 'config.json' );
```

I don't see inherent problems needing the `path.join()` but I would suggest it's better to go all-in or not at all so we don't introduce inconsistency

resource/build-scripts/build-config-file.js
```diff
@@ -1,15 +1,29 @@
...
+
+const { NODE_ENV } = process.env;
+let envConfig = {};
+
+try {
+	envConfig = require( path.join( CONFIG_SRC_DIR, `config-${NODE_ENV}.json` ) );
```
dmsnell: these should not be `NODE_ENV`

currently I see `base`, `test`, `release`, `dev`, and `updater` config files, none of which should ever be the value of `NODE_ENV`

Makefile
```diff
@@ -4,52 +4,88 @@ else
...
+	@if [ $(CONFIG_ENV) = "release" ] && [ ! -f $(CALYPSO_DIR)/config/secrets.json ]; \
+	then { \
+		if [ -z "${CIRCLECI}" ]; \
+			then { \
+				echo "$(RED)x calypso/config/secrets.json not found. Required file, see docs/secrets.md$(RESET)"; \
+				exit 1; \
+			} \
+		fi; \
+	} \
+	fi;
```
dmsnell: so it looks like we copied a BASH script into Make without transforming the paradigm. just pointing this out even though it was probably existing…

```Makefile
calypso/config/secrets.json:
ifndef $(CIRCLECI)
	$(info $(RED)x $@ not found. Required file, see docs/secrets.md$(RESET))
else
endif
```

^^^ doing that we can then incorporate it into some `make release` command or something like it.

this is a random comment, please don't feel pressured

Makefile
```diff
@@ -4,52 +4,88 @@ else
...
+
+CONFIG_ENV = 
+CALYPSO_ENV = desktop
+NODE_ENV = production
+BUILD_PLATFORM = mwl
+
+# Build sources
+# TODO: run tasks parallel when in dev mode
+build-source: checks build-config build-calypso build-desktop
+	@echo "$(CYAN)$(CHECKMARK) All parts built$(RESET)"
```
dmsnell: while `@echo` instantiates a shell the `$(info …)` function doesn't, so we can do this which is somewhat trivial and meaningless for overall performance, but is a healthy `Makefile` practice…

```Makefile
build-source: checks build-config build-calypso build-desktop
	$(info $(CYAN)$(CHECKMARK) All parts built$(RESET))
```

Makefile
```diff
@@ -4,52 +4,88 @@ else
...
+
+# Start app
+start: build-source
+	@echo "$(CYAN)Starting app...$(RESET)"
+
+	@npx electron .
+
+# Build config
+build-config: 
+	@node $(THIS_DIR)/resource/build-scripts/build-config-file.js $(CONFIG_ENV)
```
dmsnell: at this point there's not much we're using our `Makefile` for since all it's doing is duplicating messaging and calling Javascript files.

```Makefile
desktop-config/config-%.json:
	$(info Config file for environment "$(CONFIG_ENV)" does not exist. Ignoring Environment.)

desktop/config.json: BASE=desktop-config/config-base.json
desktop/config.json: ENV=desktop-config/config-$(CONFIG_ENV).json
desktop/config.json: desktop-config/config-base.json desktop-config/config-$(CONFIG_ENV).json
	@node -e "var base = require('$(BASE)'), env = require('$(ENV)'); console.log( Object.assign( base, env ) )" > $@
```

with this reconstruction `make` will only rebuild the configuration if either the base or the env-specific config has changed, and we don't have to call for it indirectly as `build-config` - we just declare our dependency as it is - a file. the node script remains minimal and while it's a long line we can see what its goal is and easily change it without worrying about reproducing the right behavior

Makefile
```diff
@@ -4,52 +4,88 @@ else
...
+
+# BUild calypso bundle
+build-calypso: 
+	@cd $(CALYPSO_DIR) && NODE_ENV=$(NODE_ENV) CALYPSO_ENV=$(CALYPSO_ENV) npm run -s build
+
+	echo "$(CYAN)$(CHECKMARK) Calypso built$(RESET)"
+
+# Build desktop bundle
+build-desktop:
+	@NODE_ENV=$(NODE_ENV) NODE_PATH=calypso/server$(ENV_PATH_SEP)calypso/client npx webpack --config ./webpack.config.js
```
dmsnell: might be helpful

```Makefile
$/=$(FILE_PATH_SEP)
$:=$(ENV_PATH_SEP)

build-desktop:
	NODE_PATH=calypso$/server$:calypso$/client
```

Makefile
```diff
@@ -4,52 +4,88 @@ else
...
+	@echo "$(CYAN)$(CHECKMARK) Config created$(RESET)"
+
+# BUild calypso bundle
+build-calypso: 
+	@cd $(CALYPSO_DIR) && NODE_ENV=$(NODE_ENV) CALYPSO_ENV=$(CALYPSO_ENV) npm run -s build
+
+	echo "$(CYAN)$(CHECKMARK) Calypso built$(RESET)"
+
+# Build desktop bundle
+build-desktop:
```
dmsnell: where `make` can really speed things up (though not really for CircleCI) is when we add in dependency information. what does the build depend on? JS files? SCSS files? assets?

```Makefile
JS_FILES=$(shell find client -type f -name "*.js")
…

build-desktop: $(JS_FILES) $(SCSS_FILES) $(ASSETS)
```

obviously that list of files is incomplete but it's not too hard to make it more complete with some options to `find`. `make` will only rebuilt the app if files have changed.

further, should we end up only needing to update some files we can use the `$?` automatic variable which returns a space-separated list of files that have changed since the target was last built

Makefile
```diff
@@ -74,141 +110,3 @@ check-node-and-npm-version-parity:
...
-build: check-node-and-npm-version-parity install
-	@echo "Building Calypso (Desktop on branch $(RED)$(CALYPSO_BRANCH)$(RESET))"
-	@cd calypso && CALYPSO_ENV=desktop MINIFY_JS=false npm run build
-	@rm -f $(THIS_DIR)/calypso/public/devmodules.*
-
-build-if-not-exists:
-	@if [ -f $(CALYPSO_JS) ]; then true; else make build; fi
-
-build-if-changed: build-if-not-exists
-	@if [ $(CALYPSO_CHANGES_STD) -eq 0 ]; then true; else make build; fi;
```
dmsnell: this and `build-if-not-exists` are already built-in to `make`, so they are slow and redundant here :smile:

alisterscott:
If we stick with CircleCI we should move to CircleCI 2.0 build syntax as 1.0 (which is currently in use) retires around the end of August. See #439

adlk:
Closing this PR in favor of #440

PR #435, opened 2018-04-05
Upgrade to webpack 4 and compat with wp-calypso

samouri:
so that wp-desktop doesn't break when we upgrade wp-calypso: https://github.com/Automattic/wp-calypso/pull/22860


samouri:
cc @astralbodies

PR #438, opened 2018-04-11
update to tip of calypso

samouri:
wp-desktop's calypso submodule is pointed to a branch that no longer exists. This happened because it was pointed to a related calypso branch while it was in progress, and then never switched back to master before merge (my bad)

I'm not really clear on the proper workflow for pointing desktop PRs to related Calypso PRs, and then what to do once the Calypso PR is merged to master. Seems funky to need to switch to a branch and then switch back to master after calypso pr merges but before merging the desktop pr.

created via:
```
cd wp-desktop
cd calypso
git checkout master
git fetch
git reset --hard origin/master
cd ..
git commit -a
``` 
cc @dmsnell


dmsnell:
I'm not sure how to evaluate this. Seems fine to me based on my ignorance.

PR #447, opened 2018-05-30
Improve spell checker

adlk:
### Description:
New implementation for the spell checker + context menu (with spelling suggestions)

### Motivation and Context:
Previously spell checking was only available for users with `en-US` as a locale. The implementation was unnecessary complicated and split over multiple files across the main and render process.

### Issues:
~~In Electron 1.8.X has introduced an issue where the spell checking APIs are not called in sub-frames.
As TinyMCE is not using the inline mode, an iframe is rendered and therefore the spell checker not properly called. This means that spell checking is currently only available in the HTML editor.~~

With Electron 1.8.4 the spell checking engine has been fixed. However dynamic generation of iframes (=everything that is happening in this web thingy) is still breaking the spell checker. As a work around, we have to downgrade to Electron 1.7.15 which luckily still gets security patches.

The spell checker issue in subframes goes up to Electron 2.x. While spell checker is completely broken in Electron 3.x. 

#### Spell checker:
1. Pull this PR, and run:
```shell
npm install # install new dependencies
npx electron-builder install-app-deps # rebuild native dependencies for electrons node version

make dev-server
```

2. After starting the dev server, run this in a second terminal:
```
make dev
```

3. Go to the editor, switch to HTML mode and you'll see misspelled words underlined.

4. Right click a misspelled word and get spelling suggestions.

#### Context Menus:
Right click on words/elements overall the app (not limited to the editor) and get context menus based on the element you've right clicked.


astralbodies:
@adlk I just changed the base on this to `develop` since we merged the `develop-next` PR. Is that okay?

adlk:
@astralbodies sure!

adlk:
@astralbodies I'm not entirely sure why it is caused (as it seems to be a bit random) but you can work around it by opening the dev tools and enter `window.location.href = "http://calypso.localhost:3000/oauth-login"` in console

astralbodies:
@adlk Can you resolve the conflicts before I test again? I tried to do it myself but the Makefile changes are complex enough that I know I'll break something if I try.

adlk:
@astralbodies all done ✅

PR #451, opened 2018-06-08
Fix webhook URL it is using the wrong one

alisterscott:
Update webhook for wp-calypso builds - this is the wrong one currently (it is using the e2e tests one)


alisterscott:
Merging this since wp-calypso PRs aren't getting the right status updates

PR #453, opened 2018-06-08
Move to Node 10.4.0 and npm 6.1.0

blowery:
Switch to node 10.4.0 and npm 6.1.0

Tracking Calypso moves

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,2 @@
 engine-strict = true
 save-exact = true
-package-lock = false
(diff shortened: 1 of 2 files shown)


sirreal:
Merging this before it landed in Calypso (https://github.com/Automattic/wp-calypso/pull/25374) seems to have broken the wp-desktop Calypso e2e tests, which now fail at the version-check step.

An example: https://circleci.com/gh/Automattic/wp-desktop/18504

astralbodies:
Gahhhhhh - I'll revert!

sirreal:
Thanks! I suspect that will fix it. It looks like the builds are pulling the latest commit from `develop`.

PR #454, opened 2018-06-08
Remove engines key in package.json

adlk:
As we are already using Calypsos .nvmrc for WP-Desktop, I've removed the engines key and replaced it with a simple check in makefile (`check-node-version-parity`) if we are using the correct nodejs version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ package:
 build: build-source package
 
 # Perform checks
-checks: check-node-and-npm-version-parity secret secret-clientid
+checks: check-node-version-parity secret secret-clientid
 
 # Check for secrets.json
 secret:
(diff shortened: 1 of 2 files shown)


adlk:
@sirreal thank for your notes. I've fixed the unary operator exception and made the version mismatch message much clearer. 

Ad adding the check to other actions: I believe we are good as it is as e.g. `build` runs `build-source` which runs `check-node-version-parity`. Plus, I would like to do some real life testing with the new makefile before we are going to optimize it further.

astralbodies:
CircleCI failed with:

> cat: calypso/.nvmrc: No such file or directory
> No .nvmrc file found

blowery:
@astralbodies in the previous step:
`Server does not allow request for unadvertised object ec8dd3b8eb488e89b4b8a1ffa97ed3ab4e73ac4c`

The submodule is now referencing a commit in calypso that's not reachable. Just need to update the submodule reference.

sirreal:
If I'm not mistaken, Node versions should be _exactly_ in sync. This repo should respect wp-calypso's. This change just removes an extra change required when updating the Node version in wp-calypso.

PR #456, opened 2018-06-11
Revert "Move to Node 10.4.0 and npm 6.1.0"

astralbodies:
Reverts Automattic/wp-desktop#453

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 engine-strict = true
 save-exact = true
+package-lock = false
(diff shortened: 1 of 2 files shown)


adlk:
@sirreal, @astralbodies I believe we should skip this in favor of #454 as pinning the node and npm version in package.json's engine key always leads to issues – especially since we are using the calypso .nvmrc. This always leads to issues during node version upgrades which now happen more often due to the switch to the 10.x track.

PR #461, opened 2018-06-22
CircleCI: Cache nvm usage

sirreal:
Cache nvm and node install on CircleCI

Should speed things up!

(Observation, not part of this PR) It would be nice to avoid the source and `nvm use` for every command:
- https://discuss.circleci.com/t/how-to-add-a-path-to-path-in-circle-2-0/11554/3
- https://discuss.circleci.com/t/sourcing-bash-file-before-every-run-command/13467

PR #462, opened 2018-06-22
CircleCI: Improve npm cache and install

sirreal:
#460 is merged into this branch (depends on package-lock)

Install with `npm ci` now that we're on recent `npm`. This aligns with Calypso and other ecosystem changes.

Cache the npm cache rather than `node_modules` which will be reinstalled with `npm ci`.
Do it consistently across jobs via references.

Should speed things up!

Closes #460

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,6 +128,7 @@ jobs:
             - calypso/server
             - calypso/config
             - calypso/package.json
+            - calypso/npm-shrinkwrap.json
             - calypso/.nvmrc
 
   win-and-linux:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,28 @@ references:
       key: nvm-0-33-11-{{ arch }}-{{ checksum "calypso/.nvmrc" }}
       paths:
         - ~/.nvm
+  npm_restore_cache: &npm_restore_cache
+    restore_cache:
+      name: Restore npm cache
+      keys:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -14,6 +14,31 @@ nvm: &nvm
...
+        - v1-npm-{{ arch }}-{{ checksum "calypso/.nvmrc" }}
+  npm_install: &npm_install
+    run:
+      name: Npm install
+      command: |
+        source $HOME/.nvm/nvm.sh
+        nvm use
+        npm ci
+        cd calypso
+        npm ci
```
sirreal: This `cd calypso; npm ci` is a bit different in that it's installing the calypso deps each time. Maybe the install steps should be unique.

sirreal: changed to match previous behavior in dd7c00a

.circleci/config.yml
```diff
@@ -92,6 +108,7 @@ jobs:
             - calypso/server
             - calypso/config
             - calypso/package.json
+            - calypso/npm-shrinkwrap.json
```
sirreal: Persist this for cache keys and npm ci

sirreal:
I could use some help here, I'm not sure why the build is failing and whether that is related.

adlk:
@sirreal i removed the not longer needed optionalDependency `appdmg` which caused this issue as it's macOS only and we are building on different OSes.

sirreal:
Rebased to fix conflicts from #461

PR #464, opened 2018-06-25
Fix RCE in Electron Windows

roundhill:
Greetings! I come from https://github.com/Automattic/simplenote-electron where I've recently fixed a vulnerability related to dragging and dropping content into electron windows that could execute remote code. This app is also similarly vulnerable, but only for secondary windows (not the main app window).

I've added a new config property that will allow us to disable drag and drop (or any window navigation action at all), and enabled it for the windows in the app that were exposed to this (About, Preferences, Secret and Wapuu). I have also disabled node integration on the `secretWindow`, because as far as I can tell it didn't need it.

**To Test**
1. Start the app, open the preferences window.
2. Drag something into the window (url, image, pdf, etc).
3. Nothing should happen 🤪
4. Verify that drag and drop still works in the post editor to add media or other acceptable content.

### DIFF
--- a/desktop/lib/window-manager/index.js
+++ b/desktop/lib/window-manager/index.js
@@ -79,6 +79,13 @@ function openWindow( windowName ) {
 			windows[windowName].handle.on( 'closed', function() {
 				windows[windowName].handle = null;
 			} );
+
+			if ( Config[settings.config].wpDragAndDropDisabled ) {
+				windows[windowName].handle.webContents.on( 'will-navigate', function( event ) {
+					event.preventDefault();
+					return false;
+				} );
+			}
 		} else {
 			settings.handle.show();
 		}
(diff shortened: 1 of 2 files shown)


desktop/lib/window-manager/index.js
```diff
@@ -79,6 +79,13 @@ function openWindow( windowName ) {
 			windows[windowName].handle.on( 'closed', function() {
 				windows[windowName].handle = null;
 			} );
+
+			if ( Config[settings.config].wpDragAndDropDisabled ) {
+				windows[windowName].handle.webContents.on( 'will-navigate', function( event ) {
```
adlk: @roundhill thanks for taking care of this 👍 

I just need to mention this for the record: This will not just disable drag and drop as it will disable all non-hash based navigation. While I don't think that this is a big deal for the windows that are currently affected, I believe we should add a note to avoid endless debugging in the future.

PR #467, opened 2018-07-02
Optimize/cleanup makefile

adlk:
- replaced `build-config-file.js` with make magic 🌈 ✨
- merged `desktop-dev` with `build-desktop` (webpack --watch is triggered by `NODE_ENV=development`) 
- merged `secret-clientid` with `secret` (now using makes `ifeq/ifneq`)
- build package without `BUILD_PLATFORM` now just builds for host system
- added OS specific file path separator
- added `CALYPSO=true/false` to `dev-server` to skip calypso server initialization.


Makefile
```diff
@@ -102,35 +109,21 @@ package:
...
+ifneq (,$(wildcard $(CALYPSO_DIR)$/config$/secrets.json))
+ifneq (43452,$(shell node -p "require('$(CALYPSO_DIR)$/config$/secrets.json').desktop_oauth_client_id"))
+	$(error "desktop_oauth_client_id" must be "43452" in $(CALYPSO_DIR)$/config$/secrets.json)
+endif
+else 
+	$(error $(CALYPSO_DIR)$/config$/secrets.json does not exist)
+endif
+
+
+CALYPSO_NODE_VERSION := $(shell cat calypso$/.nvmrc)
```
dmsnell: uh…can we not depend on `nvm` here? how about `node -v`?

Makefile
```diff
@@ -82,15 +91,13 @@ calypso-dev:
...
-
-	@echo "$(CYAN)$(CHECKMARK) Desktop built$(RESET)"
+ifeq ($(NODE_ENV),development)
+	@echo "$(CYAN)$(CHECKMARK) Starting Desktop Server...$(RESET)"
+endif
 
-# Build and watch desktop scripts
-desktop-dev:
-	@echo "$(CYAN)Starting Desktop Server...$(RESET)"
+	@NODE_ENV=$(NODE_ENV) NODE_PATH=calypso$/server$(ENV_PATH_SEP)calypso$/client CALYPSO_SERVER=true npx webpack --config $(THIS_DIR)$(/)webpack.config.js
```
dmsnell: any specific reason for `$(/)` here instead of `$/`?

Makefile
adlk: `$(shell $(NODE) -v)` wouldn't do much as we need the pinned version instead of the currently used version.
I'm happy to grab the info from calypsos package.json. It shouldn't become an issue as long as we always pin the exact version and not do semver ranges.

Makefile
adlk: This must have slipped through 😇 Thanks!

Makefile
```diff
@@ -40,33 +47,35 @@ start:
...
 	$(MAKE) start NODE_ENV=$(NODE_ENV) DEBUG=$(DEBUG)
 
-# Build config
-build-config: 
-	@node $(THIS_DIR)/resource/build-scripts/build-config-file.js $(CONFIG_ENV)
+desktop-config$/config-%.json:
+	$(info Config file for environment "$(CONFIG_ENV)" does not exist. Ignoring Environment.)
 
-	@echo "$(CYAN)$(CHECKMARK) Config created$(RESET)"
+.PHONY: desktop$/config.json
```
dmsnell: as we discussed in Slack we can refine this further by creating a kind of "lock file" for the environment variables. I don't think it's necessary to do that here in this PR but I'm leaving a note for the future reader that by writing more idiomatic `Makefile` code like is happening here (instead of using a custom JS runner) then we make it easier to refine and optimize our build process.

someone should be able to come around to this line and say, "oh hey, we don't have to regenerate the config files every time, we can add `.env` as a dependent and skip some work" whereas with the JS runner it's number one _hidden from sight_ and number two it _has_ to run every time

Makefile
```diff
@@ -40,33 +47,35 @@ start:
...
 dev: NODE_ENV = development
 dev: DEBUG = desktop:*
 dev: 
 	$(MAKE) start NODE_ENV=$(NODE_ENV) DEBUG=$(DEBUG)
 
-# Build config
-build-config: 
-	@node $(THIS_DIR)/resource/build-scripts/build-config-file.js $(CONFIG_ENV)
+desktop-config$/config-%.json:
+	$(info Config file for environment "$(CONFIG_ENV)" does not exist. Ignoring Environment.)
```
dmsnell: this one may have a bug and I know I suggested the code which means I introduced it.

this will run whenever the config file changes as well as when it's missing. this might be
a good time to invoke a shell to confirm the presence of the file. I think the following
should work (but it's probably wrong)

```Makefile
desktop-config$/config-%.json:
	@if [ -z $@ ]; then echo "Config file for environment \"$(CONFIG_ENV)\" does not exist. Ignoring environment."; fi
```

adlk: For simplicity I merged that with `desktop-config/config.json` as it failed in the JS liner as well.

Makefile
```diff
@@ -109,32 +109,18 @@ package:
...
-		fi; \
-	} \
-	fi;
-
-
-CLIENT_ID := $(shell node -p "require('$(CALYPSO_DIR)$(/)config$(/)secrets.json').desktop_oauth_client_id")
-
-# Confirm proper clientid for production release
-secret-clientid:
-	@if [ "$(CONFIG_ENV)" = "release" ] && [ ! $(CLIENT_ID) = "43452" ]; \
```
ockham: Dropping the `[ "$(CONFIG_ENV)" = "release" ]` criterion now causes `make dev-server` to fail with 

```
Makefile:123: *** "desktop_oauth_client_id" must be "43452" in /home/bernie/src/wp-desktop/calypso/config/secrets.json.  Stop.
```

when using and oauth ID different from the official WP.com desktop app, even tho docs still [suggest](https://github.com/Automattic/wp-desktop/blob/develop/docs/secrets.md#calypso-secrets) it's fine to use any generated app secrets.

So I think we either need to add the criterion back, or update our docs.

@adlk

ockham: (Filed #483 to track this.)

PR #469, opened 2018-07-04
Cache calypso

adlk:
### Description:
Introduce a caching strategy for calypso bundles as they can be shared between wp-desktop builds.

### Motivation and Context:
Every time we make changes with the desktop app, we do a complete calypso rebuild that takes 5+ minutes even though calypso did not change at all for at least 95% of all builds. (Not counting e2e)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ references:
       paths:
         - ~/.npm
   calypso_cache_paths: &calypso_cache_paths
-    - calypso-cached-hash
+    - calypso-hash
     - build
     - resource/certificates
     - calypso/public
@@ -59,7 +59,6 @@ references:
     run:
       name: Prepare calypso cache
       command: |
-              # touch calypso-cached-hash
               git rev-parse @:./calypso > calypso-current-hash
   calypso_restore_cache: &calypso_restore_cache
     restore_cache:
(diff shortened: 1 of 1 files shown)


ockham:
I've added @blowery as a reviewer since he's one of the people that knows our bundling most intimately.

dmsnell:
also calling in @jsnajdr since this hints at a bigger project.

I'd _love_ to have CircleCI build Calypso as an artifact. we could also have `dserve` load it instead of build it too. we could have [ICFY](iscalypsofastyet.com) lod it too instead of building it. so many possibilities!

that being said let's coordinate a plan so we don't duplicate our efforts

adlk:
@dmsnell we should definitely think about those options long term. The question is if this isn't a great interim solution until we have something "better".

.circleci/config.yml
```diff
@@ -45,6 +45,43 @@ references:
...
+    - calypso/package.json
+    - calypso/npm-shrinkwrap.json
+    - calypso/.nvmrc
+  calypso_prepare_cache: &calypso_prepare_cache
+    run:
+      name: Prepare calypso cache
+      command: |
+              if [ -n "${CALYPSO_HASH}" ]; then
+                # cache bust e2e tests
+                uuidgen > calypso-current-hash
```
sirreal: Why use a unique ID not depending on the Calypso hash? Couldn't we just `echo "${CALYPSO_HASH}" > calypso-current-hash`?

PR #471, opened 2018-07-09
Extract nvm setup to script

sirreal:
Extract `nvm` boot logic to a reusable script.
Must be `source`d to expost the `NVM_DIR` environment variable.


sirreal:
I think it would be better to set `NVM_DIR` via yaml and avoid needing to source the new file.

PR #477, opened 2018-07-16
Upgrade Babel to 7.0.0 beta 53

jsnajdr:
This patch upgrades Babel to 7.0.0 beta 53 and adds a `babel.config.js` file.

The config file uses Babel Env Preset to specify a compilation target (Electron) and is used to compile the `wp-desktop` files and also the `calypso/server` ones that are included by `desktop/server/server.js`.

With `babel.config.js`, Babel will no longer load `calypso/.babelrc.js`, so the desired config needs to be specified by the parent project.

This patch is a prerequisite to upgrading Babel in Calypso, too (Automattic/wp-calypso#26020)

**How to test:**
The risk is that after the upgrade the build process will output some JS syntax or use some JS builtins that are not yet supported by Electron. Therefore, almost anything can break.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,9 @@
+module.exports = {
+	presets: [
+		[ '@babel/env', { targets: { electron: '1.8.4' } } ],
+		'@babel/react'
+	],
+	plugins: [
+		'@babel/plugin-proposal-class-properties'
+	]
+};
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,28 +46,131 @@
       }
     },
     "@babel/core": {
-      "version": "7.0.0-beta.44",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.0.0-beta.44.tgz",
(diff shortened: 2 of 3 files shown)


ockham:
> @ockham On the app run by `make start`?

I ran `make dev-server` in one terminal and then `make dev` in another, as suggested by p4TIVU-906-p2. I can try `make start` tho.

> Is there any chance of seening the devtools console on the running app?

Yep, good call! Using Ctrl+Shift+I, I see 

> public-api.wordpress.com/rest/v1.1/me?meta=flags&<REDACTED> Failed to load resource: the server responded with a status of 405 ()
> Failed to load resource: the server responded with a status of 405 ()

and

> XMLHttpRequest cannot load https://public-api.wordpress.com/rest/v1.1/me?meta=flags&<REDACTED>. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://calypso.localhost:3000' is therefore not allowed access. The response had HTTP status code 405.

ockham:
Got this to work by adding `http://calypso.localhost:3000` to both 'Redirect URLs' and 'JavaScript Origins', (contrarily to what [the docs](https://github.com/Automattic/wp-desktop/blob/develop/docs/secrets.md) currently suggest)

image

PR #479, opened 2018-07-20
Bring webpack plugins section in sync with wp-calypso

adlk:
### Description

Webpack plugins in `webpack.shared.js` should be in sync with the calypso webpack config.

### Additional Note

As webpack is using the `browser` key from the package.json of dependencies for the `web` target, we run into the issue that the module `matches-selector` can't be found as it's actually `component-matches-selector` for non-web targets.

We can work around that issue by adding the following line but it might be a common trap to fall into in the future.
```js
new webpack.NormalModuleReplacementPlugin( /^matches-selector$/, 'component-matches-selector' ), // Required as this module is not compiled for browser target

```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2343,6 +2343,11 @@
       "integrity": "sha1-+PeLdniYiO858gXNY39o5wISKyw=",
       "dev": true
     },
+    "buffer-from": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/buffer-from/-/buffer-from-1.1.0.tgz",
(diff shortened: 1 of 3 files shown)


jsnajdr:
Could we also merge _webpack.shared.js_ and _webpack.config.js_ into one file? Unless I'm missing something, the _shared_ config is not actually shared by anything.

adlk:
@jsnajdr I believe we could do so. 

I want to compile the files in `public_desktop` as well which results in multiple entries/output files. I will check what's possible and try to do it then.

ockham:
> Could we also merge _webpack.shared.js_ and _webpack.config.js_ into one file? Unless I'm missing something, the _shared_ config is not actually shared by anything.

:100: on cutting down on all those ~duplicated~ ~triplicated~ multiple configs

PR #481, opened 2018-07-30
Fix calypso cache platform dependency

adlk:
This PR is fixing an issue in #469 (reverted) as the CircleCI cache key was `arch` dependent. This however results in calypso not being found in the linux job.

Dropping `arch` from the cache key makes the cached calypso version available for all platforms.


adlk:
I was mislead by something I broke in another branch 🤦‍♂️ 
Everything is in order.

PR #485, opened 2018-08-05
Fix: Make dev-server doesn't use `CONFIG_ENV`

### LINKED ISSUE #484: `make dev-server` and `make dev` give blank window
#### Steps to reproduce
1.`make dev-server` (with `calypso/config/secrects.json` using the [official desktop app](https://developer.wordpress.com/apps/43452/)) secrets to work around #483 
2. In another terminal, `make dev`

#### What I expected
The app to start & work

#### What happened instead
App started a blank window

#### App Version / OS version

`develop` branch / Ubuntu 16.04

#### Screenshot / Video

image

#### Context / Source
Nothing suspicious in either console.

adlk:
### Description:
The make `dev-server` target isn't correctly passing `CONFIG_ENV` to the `desktop/config.json` target. When running `make dev` this results in the app rendering just a white screen as it's looking for `http://127.0.0.1:41050` instead of `http://calypso.localhost:3000`.

### Motivation and Context:
This PR closes #484

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,13 +47,15 @@ start:
 dev-server: CONFIG_ENV = development
 dev-server: CALYPSO_ENV = desktop-development
 dev-server: NODE_ENV = development
-dev-server: checks desktop$/config.json
+dev-server: checks
 	@echo "\n\n$(GREEN)+------------------------------------------------+"
 	@echo "|                                                |"
 	@echo "|    Wait for calypso to start the dev server    |"
 	@echo "|       and start the app with \`make dev\`        |"
 	@echo "|                                                |"
 	@echo "+------------------------------------------------+$(RESET)\n\n"
+
+	$(MAKE) desktop$/config.json CONFIG_ENV=$(CONFIG_ENV)
 
 	@npx concurrently -k \
 	-n "Calypso,Desktop" \


Makefile
```diff
@@ -47,14 +47,16 @@ start:
...
-dev-server: checks desktop$/config.json
+dev-server: checks
 	@echo "\n\n$(GREEN)+------------------------------------------------+"
 	@echo "|                                                |"
 	@echo "|    Wait for calypso to start the dev server    |"
 	@echo "|       and start the app with \`make dev\`        |"
 	@echo "|                                                |"
 	@echo "+------------------------------------------------+$(RESET)\n\n"
 
+	$(MAKE) desktop$/config.json CONFIG_ENV=$(CONFIG_ENV)
```
ockham: This fixes the issue, but it goes kind of against `make`'s purpose to call it directly here, losing the [target/dependencies semantics](https://github.com/Automattic/wp-desktop/pull/485/files#diff-b67911656ef5d18c4ae36cb6741b7965L50). Is there any way to pass on env vars to dependent targets?

ockham:
Rebased.

Makefile
adlk: Yes, you are right.
After spending hours on #467 to tackle this, I've decided to use this shortcut to get it to a working state as we (@dmsnell and I) ran in to a couple issues – same you are now seeing on #487. 

I'd however love to see this done right by using makes magic to the fullest!

PR #487, opened 2018-08-06
Makefile: Pass CONFIG_ENV to desktop.json build target

### LINKED ISSUE #484: `make dev-server` and `make dev` give blank window

ockham:
### Description:
Alternative (purportedly more semantic) approach to #485 

The make `dev-server` target isn't correctly passing `CONFIG_ENV` to the `desktop/config.json` target. When running `make dev` this results in the app rendering just a white screen as it's looking for `http://127.0.0.1:41050` instead of `http://calypso.localhost:3000`.

### Motivation and Context:
This PR closes #484


ockham:
Rebased.

sirreal:
I've been unable to test: p1533572783000370-slack-wordpress-desktop

Makefile
```diff
@@ -66,11 +66,9 @@ dev: DEBUG = desktop:*
...
 
-
-BASE_CONFIG := $(THIS_DIR)/desktop-config/config-base.json
-ENV_CONFIG := $(THIS_DIR)/desktop-config/config-$(CONFIG_ENV).json
-
 .PHONY: desktop$/config.json
+desktop$/config.json: BASE_CONFIG = $(THIS_DIR)/desktop-config/config-base.json
+desktop$/config.json: ENV_CONFIG = $(THIS_DIR)/desktop-config/config-$(CONFIG_ENV).json
 desktop$/config.json:
 ifeq (,$(wildcard $(ENV_CONFIG)))
```
adlk: @ockham I initially had this approach in #467 but it turned out that `ifeq (,$(wildcard $(ENV_CONFIG)))` never evaluates to true.

Running this running results in:
```console
adlk@work: $ make build CONFIG_ENV=release
Makefile:73: Config file for environment "release" does not exist. Ignoring environment.
```

ockham:
Okay, I gave this another shot.

I tried to consider what we want to do here, and align that with what would be a semantic way to do this in a `Makefile`. `Makefile`s are really quite heavily based on targets and dependencies, and both are normally filenames. So rather than passing variables down a chain, a top-level target (such as `make-dev` will ideally require a specific dependent filename that includes what is now an argument.

Makefile
```diff
@@ -66,20 +67,10 @@ dev: DEBUG = desktop:*
...
 	$(MAKE) start NODE_ENV=$(NODE_ENV) DEBUG=$(DEBUG)
 
-
-BASE_CONFIG := $(THIS_DIR)/desktop-config/config-base.json
-ENV_CONFIG := $(THIS_DIR)/desktop-config/config-$(CONFIG_ENV).json
-
-.PHONY: desktop$/config.json
-desktop$/config.json:
-ifeq (,$(wildcard $(ENV_CONFIG)))
-	$(warning Config file for environment "$(CONFIG_ENV)" does not exist. Ignoring environment.)
```
ockham: I've removed this clause, since `desktop-config/config-%.json` is now a dependent target, so `make` will check for us if the file really exists.

Makefile
```diff
@@ -66,20 +67,10 @@ dev: DEBUG = desktop:*
...
-else
-	$(eval EXTENDED = true)
-endif
-	@node -e "const base = require('$(BASE_CONFIG)'); let env; try { env = require('$(ENV_CONFIG)'); } catch(err) {} console.log( JSON.stringify( Object.assign( base, env ), null, 2 ) )" > $@
+desktop$/merged-config-%.json: desktop-config$/config-base.json desktop-config/config-%.json
+	@node -e "const base = require('$(THIS_DIR)/$<'); let env; try { env = require('$(THIS_DIR)/$(word 2,$^)'); } catch(err) {} console.log( JSON.stringify( Object.assign( base, env ), null, 2 ) )" > $@
```
ockham: `$<` is the first dependency (`desktop-config$/config-base.json`), `(word 2,$^)` is the second (`desktop-config/config-%.json`). This reads a bit Greek -- maybe we can alias them inside the rule (though at first try, I wasn't able to).

Makefile
```diff
@@ -47,14 +47,15 @@ start:
...
-dev-server: checks desktop$/config.json
+dev-server: checks desktop$/merged-config-development.json
 	@echo "\n\n$(GREEN)+------------------------------------------------+"
 	@echo "|                                                |"
 	@echo "|    Wait for calypso to start the dev server    |"
 	@echo "|       and start the app with \`make dev\`        |"
 	@echo "|                                                |"
 	@echo "+------------------------------------------------+$(RESET)\n\n"
 
+	cp desktop$/merged-config-$(CONFIG_ENV).json desktop$/config.json
```
ockham: Copying this manually as part of this target is still somewhat suboptimal (but a consequence of not being able to pass around target variables).

Makefile
```diff
@@ -34,7 +34,8 @@ TEST_PRODUCTION_BINARY = false
 
 # Build sources
 # TODO: run tasks parallel when in dev mode
-build-source: checks desktop$/config.json build-calypso build-desktop
+build-source: checks desktop-config$/config-base.json build-calypso build-desktop
+	cp desktop-config$/config-base.json desktop$/config.json
```
ockham: We're not merging any env specific config into the base config here, so we don't need to go through the `desktop$/merged-config-$(CONFIG_ENV).json` target.

PR #503, opened 2018-09-20
Ignore SASS/CSS imports when building desktop

jsnajdr:
Server build doesn't need to process any CSS imports -- that's only for Calypso client build. Fixed build errors in Automattic/wp-calypso#26820.


flootr (approved):
LGTM

PR #515, opened 2018-10-19
Testing: Adding some e2e tests for mac application

bsessions85:
### Description:
This adds e2e tests for Log-in, Create Post, and Log-out. The tests will run in CircleCI in the e2e tests step in the mac environment.

### Motivation and Context:
To improve quality we want to be able to test some basic functionality on build

### How Has This Been Tested:
This has been tested by running in CircleCI. The e2e Tests step shows the test results.


package.json
```diff
@@ -63,6 +69,7 @@
     "express": "4.15.3",
     "lodash.clonedeep": "4.5.0",
     "lodash.debounce": "4.0.8",
+    "nan": "2.11.1",
```
bsessions85: Adding this was the only way I could work around some build issues with errors about 'unresolved node modules'

rachelmcr:
Very excited about these tests! It looks like we're going to end up with quite a bit of duplication of components and page objects between this and the https://github.com/Automattic/wp-e2e-tests repo. Are there already ideas or plans for how to avoid that duplication or to handle maintenance for those going forward?

cc @loremattei I just want to make sure you're aware of these tests, in case you have any feedback at this stage. :)

bsessions85:
> Are there already ideas or plans for how to avoid that duplication or to handle maintenance for those going forward?

@rachelmcr  That is a qood question. At this point I think they should probably be kept completely separate, mostly because the desktop app uses a specific version of Calypso and isn't always changing. Trying to link them somehow would cause problems since things change much slower for desktop and tests would break as things are changed in wp-calypso. Eventually it might be nice to do some versioning on e2e tests or even just the page objects so that they are easier to pull in to projects like this.

loremattei:
@rachelmcr thank you very much for the ping! I'm very excited about these test too! 
I agree that it'd be very nice to be able to sync the tests when we update the Desktop app to a new version of Calypso (~ every 3 weeks).

PR #547, opened 2018-12-17
Remove EMIT_STATS from desktop tests

sirreal:
Added in a5d87d2c / #539 

I suspect this is slowing down Calypso Desktop tests significantly, however it's hard to tell because of the way Desktop/Calypso tests are set up.

@blowery do we need the stats?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -199,7 +199,7 @@ jobs:
                       CONFIG_ENV=test
                     fi
 
-                    make build-source CONFIG_ENV=$CONFIG_ENV EMIT_STATS=true
+                    make build-source CONFIG_ENV=$CONFIG_ENV
                   else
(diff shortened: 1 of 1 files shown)


blowery:
@sirreal I think we're only using EMIT_STATS to get the progress info.

sirreal:
Do you think the extra work is slowing us down here?

https://github.com/Automattic/wp-calypso/blob/8632b8bc5c4394b7d03d6b148f174829f3ffac30/webpack.config.js#L358-L370

If we remove the stats and progress, I'd hope the builds are faster. We'll likely need to increase the timeout on the build step. It's at 20m right now, but I've noticed some builds >30m.

jtreanor:
`EMIT_STATS` was disabled as part of https://github.com/Automattic/wp-desktop/pull/564. Along with the other changes, it seems to have given a modest speedup.

sirreal:
Superseded by #564

PR #557, opened 2019-01-29
CI: Add terser and babel cache cache

sirreal:
Applies the same caching the Calypso proper uses for terser and babel cache to improve build times.

I'm not sure how to test this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,42 @@
 version: 2
 
 references:
+  setup-test-dirs: &setup-test-dirs
+    name: Create Directories for Results and Artifacts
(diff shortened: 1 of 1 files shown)


sirreal:
> since the desktop app bundles all of the js into one file, I'm not sure the terser cache is going to be much help...

Do you mean that the desktop app does not code split and that terser runs after things are bundled into a single chunk; therefore we'll never have a cache hit unless it's the exact same build which is already handled elsewhere?

blowery:
Yup, that exactly

sirreal:
Thanks for the insight 👍

PR #562, opened 2019-02-11
Mirror move to file refs in calypso

blowery:
Companion to https://github.com/Automattic/wp-calypso/pull/30681

Moves to using `npm ci` to install calypso in one place and drops NODE_ENV usage when triggering calypso build steps that may kick off an install. NODE_ENV=production skips the install of devDependencies, which calypso needs to build itself.


blowery:
Not sure why this is failing...

PR #564, opened 2019-02-22
CircleCI: break 'build' job into macOS job for releases and Linux job for Calypso

jtreanor:
### Description:

This updates the CircleCI config to allow Calypso PR builds to run on Linux.

Note that the config actually has very few changes. The large diff is due to the yaml indentation change to make it re-usable.

It is the companion PR to https://github.com/Automattic/wp-calypso/pull/30938.

### Motivation and Context:

The `build` job of wp-desktop's CircleCI config is run many times a day on calypso PRs. Currently, it runs on a macOS image. This is problematic as it eats into the majority of our CircleCI macOS plan unnecessarily.

Unfortunately, it is not as simple as just changing it to run on Linux as that causes the calypso build to crash with an out of memory error.

The solution I have come up with it to break the `build` job into two separate jobs:

- `build`: This is the job that runs on Calypso PRs. It runs on Linux with JS minifying disabled. This allows it to run without memory errors. It has a bonus of making these PR builds quite a bit faster.
- `build-release`: This is the job that only happens on changes to wp-desktop. It does a full release build of calypso.

The workflow now looks like this:

image

### How Has This Been Tested:

This has been tested both by running on CircleCI (see checks on this PR) and by building locally with `make build`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,7 @@ references:
     restore_cache:
       name: Restore calypso cache
       keys:
-        - v5-calypso-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{ checksum "calypso-current-hash" }}
+        - v6-calypso-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{ checksum "calypso-current-hash" }}
   calypso_finalize_cache: &calypso_finalize_cache
     run:
       name: Finalize calypso cache
@@ -88,7 +88,7 @@ references:
   calypso_save_cache: &calypso_save_cache
     save_cache:
       name: Save calypso cache
-      key: v5-calypso-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{ checksum "calypso-current-hash" }}
+      key: v6-calypso-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{ checksum "calypso-current-hash" }}
       paths: *app_cache_paths
   webhook_notify_success: &webhook_notify_success
     run:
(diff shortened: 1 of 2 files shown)


jtreanor:
I have updated the calyso ref to the merge commit of https://github.com/Automattic/wp-calypso/pull/30938.

jtreanor:
@blowery Great, thanks. I will leave it open for now so they can take a look.

jtreanor:
After a little more testing there is a failure in `build` when `CALYPSO_HASH` is set. I'll make a small change to fix that and update here when its done.

.circleci/config.yml
```diff
@@ -1,12 +1,12 @@
-version: 2
+version: 2.1
```
sirreal: Is there a reason for 2.1? Is that the version that includes the feature I believe they call "build processing", which is essentially parameterized jobs? Are we using any of those features?

**Edit** Oh, I see we're using `orbs` now 😮

.circleci/config.yml
```diff
@@ -135,114 +141,117 @@ references:
...
-          command: |
-                  # only use the updater when we are building from master or a release branch
-                  # if [[ $CIRCLE_BRANCH =~ ^release(.*)|(^master$) ]]
-                  # then
-                  #   CONFIG_ENV=release
-                  # fi
 
-                  # TODO: update accordingly when code from above changes
-                  CONFIG_ENV=release
+orbs:
```
sirreal: > Orbs are packages of config that you can use to quickly get started with the CircleCI platform. Orbs enable you to share, standardize, and simplify config across your projects.

[Using Orbs](https://circleci.com/docs/2.0/using-orbs/)

Are we intending to share this at all, or is to reuse a set of steps?

For a locally reusable set of steps, we're using commands in wp-calypso:

https://github.com/Automattic/wp-calypso/blob/d28aab8c217669f1d3fbf7a9286df7506e3fec03/.circleci/config.yml#L144-L151

```yaml
commands:
  prepare:
    description: "Prepare Job"
    steps:
      # folders to collect results
      - run: *setup-results-and-artifacts
      - attach_workspace:
          at: '~'

# …

jobs:
  some-job:
    steps:
      - prepare
```

Would that be a better fit, or is there other motivation for using orbs?

.circleci/config.yml
```diff
@@ -135,114 +141,117 @@ references:
...
-                  set +e
+jobs:
+  build:
+    docker:
+      - image: circleci/node:10.14.0-browsers
+    environment:
+      CONFIG_ENV: test
+      MINIFY_JS: false
+      NODE_ARGS: --max_old_space_size=2048
+    steps:
```
sirreal: Is this missing from this job?
```suggestion
    working_directory: ~/wp-desktop	
    steps:
```

.circleci/config.yml
jtreanor: Yes, exactly.

If there is a reason not to use the newer version I could refactor the config, but this makes it simpler.

.circleci/config.yml
jtreanor: Oh, I didn't realize you could declare commands directly without an inline Orb. That seems a better fit.

Of course this still requires using CircleCI 2.1.

.circleci/config.yml
jtreanor: 👍  I guess since we specify this in every other job its no harm to be specific. Without it it defaults to `~/project` which works fine but consistency is better.

.circleci/config.yml
jtreanor: Done. As a bonus, this makes the diff easier to read since there is now no indentation change.

jtreanor:
@sirreal Thanks for your comments. Yes, it is ready for review. I spotted the `CALYPSO_HASH` issue this morning but that is now fixed. Sorry for the confusion about that.

.circleci/config.yml
```diff
@@ -211,38 +198,59 @@ jobs:
...
+      - build-and-test
+      - persist_to_workspace:
+          root: ~/wp-desktop
+          paths: *app_cache_paths
+
   linux:
     docker:
-      - image: electronuserland/builder:wine-mono
-    working_directory: /wp-desktop
+      - image: circleci/node:10.14.0-browsers
```
sirreal: Harmless, but this build uses `nvm` to get the node version Calypso wants. We could likely use a lighter image.

jtreanor: Yes, we possibly could but the reason I am using this image is because of the `-browsers` functionaltiy which enable us to headless browser testing. Without this, we would have to do lots of boilerplate configuration ourselves.

jtreanor:
Thanks, @sirreal! I'll see who the best person for that approval is.

Thankfully, apart from the refactor, the release path is unchanged here so this PR shouldn't have packaging/release implications.

jtreanor:
Conflicts with `develop` are resolved and this is ready for a final review 🤞

jtreanor:
Thanks, @loremattei!

sirreal:
Thanks for the work here @jtreanor! Slow desktop tests have been a pain in Calypso development for a while, so I'm happy to see movement here.

PR #570, opened 2019-02-27
CircleCI: Don't trigger build on push to test branches

jtreanor:
### Description:

In order to test Calypso PRs, branches with the format `tests/` are pushed to wp-desktop. To stop the push building as well as the API, yesterday we enabled "Only Build Pull Requests" on the project.

This is a small config change to exclude these branches from being triggered on pushes. It doesn't prevent API triggered builds on test branches (see [this](https://circleci.com/gh/Automattic/wp-desktop/33919) build).

It will allow us to disabled "Only Build Pull Requests" so that our release workflow is uninterrupted (we can build on tags and release branches).

### How Has This Been Tested:

I have mirrored these changes to [`tests/dont-trigger`](https://github.com/Automattic/wp-desktop/tree/tests/dont-trigger). I tested that pushes to these branches do not appear [in the CircleCI UI](https://circleci.com/gh/Automattic/wp-desktop/tree/tests/dont-trigger) and that API builds can still be triggered (see [this](https://circleci.com/gh/Automattic/wp-desktop/33919) build).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -410,10 +410,14 @@ workflows:
     jobs:
       - build:
           filters:
+            branches:
+              ignore: /tests.*/
             tags:
               only: /.*/
       - build-release:
           filters:
+            branches:
+              ignore: /tests.*/
             tags:
               only: /.*/
       - windows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -411,13 +411,13 @@ workflows:
       - build:
           filters:
             branches:
-              ignore: /tests.*/
+              ignore: /tests\/.*/
             tags:
               only: /.*/
       - build-release:
           filters:
             branches:
-              ignore: /tests.*/
+              ignore: /tests\/.*/
             tags:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -410,10 +410,14 @@ workflows:
     jobs:
       - build:
           filters:
+            branches:
+              ignore: /tests.*/
```
blowery:
```suggestion
              ignore: /^tests\//
```
might be safer? right now it'll match any branch with `tests` in the name

jtreanor: Actually, the `^` is not needed here. The current regex matches branches that start with `tests` (I've tested the behaviour on CircleCI).

I agree that adding `\/` makes sense. I'll do that now.

jtreanor: Updated now.

I have verified that branches containing `tests/` not at the start are not ignored by this.

blowery: interesting! Thanks for testing.

PR #580, opened 2019-03-29
Revert "CircleCI: Use mirror for downloading Electron binaries"

jtreanor:
Reverts Automattic/wp-desktop#579

The mirror seems to cause intermittent failures like this 😞 

```
Error: Generated checksum for "electron-v1.7.16-linux-x64.zip" did not match expected checksum.
```

However, Github seems to be working again, so lets revert this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,10 +26,6 @@ references:
         nvm install "$NODE_VERSION"
         nvm alias default "$NODE_VERSION"
         nvm use "$NODE_VERSION"
-
-        # Use an Electron mirror to avoid issues downloading Electron binaries from Github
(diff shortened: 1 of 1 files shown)


blowery:
Just for reference, I had a support ticket with Circle (48787) and got this response:

> Thanks for reporting this to us. We're not sure at this moment why is happening, but in our tests, upgrading to node 11 resolved the problem. Hence in the meantime, using circleci/node:11 or running a command to upgrade the node installation to node 11 could be a workaround for the issue. We will update you again once we find a fix for this.

Though if github is working again... yay?

jtreanor:
Thanks for the context @blowery!

Hopefully things have stabilised but otherwise we can try something like the have suggested.

Issue #582, opened 2019-04-04
Block Editor: Closing and re-opening the app with the editor open leads to a login page


rachelmcr:
h/t @pauljacobson for the original report.

#### Steps to reproduce
1. Start and save a new post in the block editor.
2. Close the app with the editor still open.
3. Re-open the app.

#### What I expected

I expected the app to re-open with the editor still open to the post I was working on. (This is how the classic editor behaves in the app.)

#### What happened instead

I see a login page.

For a public site, I am taken to the regular Calypso login page (not the desktop-specific login page):

![Screenshot 2019-04-04 19 21 19](https://user-images.githubusercontent.com/8658164/55579201-9eb07080-570f-11e9-9313-e5168169dd41.png)

For a private site, I am taken to this private site login page:

![image](https://user-images.githubusercontent.com/8658164/55579251-b851b800-570f-11e9-828e-710ecf5ed4ba.png)

In addition, the app ends up in a half logged-in/logged-out state. You can use the links on those login pages to view the logged-out WordPress.com homepage and other logged out pages. You can use the Help button in the bottom right to access logged-in Calypso again, but even then trying to start a new post in the block editor leads to a login page.

#### App Version / OS version

WPDesktop 4.1.0-beta1

rachelmcr:
I ran into a new and exciting variation of this issue today:

1. Log out of the app.
2. Log in again.
3. Go to My Sites > Blog Posts.
4. Select a post to open in the block editor.

Result: It opens a login screen instead.

Quitting and restarting the app at any point seems to resolve the issue (i.e. quitting the app before logging in again prevents it from appearing, or quitting after you see the issue resolves it). But following the steps above without restarting the app consistently reproduces the issue for me, regardless of the site or post I choose.

blowery:
Seems like when we restore the app and try to load Gutenberg, we're using the same frame nonce every time instead of generating a new one...

blowery:
Looks like we may need to enhance https://github.com/Automattic/wp-calypso/blob/master/client/gutenberg/editor/calypsoify-iframe.tsx#L433 somehow.

blowery:
@rachelmcr's second repro seems to be a similar problem: a stale frame nonce.

rachelmcr:
This came up again in beta testing version 4.3.0-beta1 (h/t @thehenrybyrd, internal ref: p5T066-V5-p2). This can lead to some really bizarre and confusing user flows, depending on what steps you take once you hit that login screen.

[...]

PR #594, opened 2019-05-07
Fix Login request on critical locations

loremattei:
This should mitigate https://github.com/Automattic/wp-desktop/issues/582.

The issue seems to be caused by a reuse of stale site nonces inside Calypso at startup and happens when WPDesktop wants to open a post in the block editor at startup. For some reasons, opening the post lists page before of the post fixes the issue. 

This PR should reverted when a proper fix gets merged into Calypso.

### DIFF
--- a/desktop/server/index.js
+++ b/desktop/server/index.js
@@ -114,7 +114,9 @@ function startServer( started_cb ) {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


loremattei:
Thanks!

PR #606, opened 2019-05-31
Build: use calypso build

sirreal:
Second take at #599. It works now 🤷‍♂


designsimply:
@sirreal I saw that @mmtr mentioned in #599 (moved to #606) that it may unblock being able to enable Gutenberg feature flags in the WPDesktop app. Is that the hope here or is this PR only about fixing problems with getting a working build?

sirreal:
> Is that the hope here or is this PR only about fixing problems with getting a working build?

Some of the errors @codebykat mentions in https://github.com/Automattic/wp-calypso/pull/32673#issuecomment-498666089 look very similar to the ones I struggled with when getting this build to work. Using more of a shared build system between Calypso and the desktop app may help.

The work in this PR is probably a good direction, but I do not have time to focus on this right now. I'd be happy if any interested parties contribute to landing it if it seems beneficial.

designsimply:
I asked about this and now I've found https://github.com/Automattic/wp-calypso/pull/33230 and https://github.com/Automattic/wp-desktop/pull/608. Do those help solve your build troubles?

sirreal:
> Do those help solve your build troubles?

If I recall correctly, this was building, so I don't think there were troubles. I don't have time to work on this right now so I haven't tested further.

Issue #651, opened 2019-07-23
Broken back button with block editor (Gutenberg)


sjaved87:
#### Steps to reproduce
1. Go to My Sites -> Site -> Posts
2. Add or edit an existing post
3. Click on the back button on top left-hand side

#### What I expected
It should return the user back to the previous page

#### What happened instead
It does nothing.

#### App Version / OS version
4.2.0
Mojave 10.14.5 (18F132)

#### Screenshot / Video
https://cld.wthms.co/e3HpLv

#### Context / Source

####Workaround for the issue I am facing is that:

Switch to the classic/old editor - screenshot -> https://cld.wthms.co/UGvpL7 -, click on the close button on the top left to go back to the previous page - screenshot -> https://cld.wthms.co/q7uqyi.

#### reported in #2200100-zen

ivan-ottinger:
User replied back sharing more info:

WordPress.com app version: 4.2.0
OS: Windows 10 (they say it is the latest version, but not much more info)

They also confirmed they are using Gutenberg and are editing existing post/page.

kspilarski:
Same issue here, with the user on **WordPress.com desktop app (4.2)**, and **Mojave 10.14**. 

- **Info:** 
In the desktop app, after publishing and/or updating the post, the left return arrow does not work and it's stuck in the same post. 

- **Follow-up ticket:** #2200927-zd 

<img width="811" alt="Blog+Posts+9+KP++WordPress com+2019-07-23+17-04-18" src="https://user-images.githubusercontent.com/36863633/61723702-df8ace80-ad6c-11e9-8af7-02c7e121df4d.png">

zdenys:
Another report  on `4.2` in 2202629-zen

rachelmcr:
I tested and reproduced the issue as described: the block editor back button doesn't do anything. It fails to close even if you try to close it immediately without making any changes.

I also opened the dev console (before opening the editor, using the shortcut `CTRL + SHIFT + i` on macOS) and the only error I see is when the editor first opens:

```
Failed to set referrer policy: The value 'strict-origin-when-cross-origin' is not one of 'no-referrer', 'no-referrer-when-downgrade', 'origin', 'origin-when-cross-origin', or 'unsafe-url'. The referrer policy has been left unchanged.
```

There are no other console errors or failed network requests when I click the back button to exit the editor.

[...]

PR #662, opened 2019-08-30
[Electron Upgrade] Upgrade Electron to v4.2.9

nsakaimbo:
### Description:
This PR upgrades Electron from v1.7.16 to v4.2.9. Electron's `v4.x` release line [embeds a custom version of Node.js v10.x](https://github.com/electron/node), which is aligned with Calypso's current Node.js version requirements.

### To Functionally Test:

Run `npm install` and then `make build` to test the built app.

### Motivation and Context:

The Electron version in wp-desktop is grossly out-of-date with respect to the latest available and therefore missing out on a number of important performance improvements and bug fixes. 

### How Has This Been Tested:

- [x] Functional smoke-testing
- [x] Existing CI test suite

### Known Issues

~~Sometimes the back button isn't responsive when trying to leave the editor view - unsure why this happens, not sure if this is a regression.~~ 
**Update**: Fixed in latest Calypso.

### Outstanding Issues: Electron-Spellchecker

`electron-spellchecker`, whether integrated directly from source or via prebuilds, only seems to work with Calypso through Electron v1.7.16, and breaks with _any_ version of Electron ≥ 1.x. (The context (right-click) menu still works, but the actual spellchecking feature no longer works.) Based on further experimentation/research, I believe the issue stems at least in part from the spellchecker not playing nicely with iframe content. 

Examining the repo's history, it seems this issue was encountered during the initial development phase and the decision was made to [downgrade Electron](https://github.com/Automattic/wp-desktop/pull/447/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L41) at that time to preserve functionality. 

I've also been made aware that the spellchecker to-date has been subject to several notable bugs since it was introduced. Fixing and improving the spellchecker might be deserving of its own dedicated project, and we might need assistance from Calypso to brainstorm some workarounds.

### Todos:

- [x] Spike out POC
- [x] Discuss and vet this approach with stakeholders
- [x] Updated dependencies:
            * Electron
            * electron-builder
            * electron-updater
            * electron-spellchecker (from prebuilds to source)
- [x] Updated the following APIs for breaking changes since v1.x:
            * [`makeSingleInstance`](https://electronjs.org/docs/api/breaking-changes#appmakesingleinstance)
            * [`crashReporter`](https://electronjs.org/docs/api/breaking-changes#crashreporter)
            * **Undocumented** breaking change in `session.setProxy` API
- [x] Unit test review
- [x] CI review/tests: fix an issue with Selenium not correctly booting the application

Before deploying this upgrade to end-users, a quality assurance pass would be great.


rachelmcr:
> Sometimes the back button doesn't responsive when trying to leave the editor view - unsure why this happens, not sure if this is a regression

That's curious. This was reported as an issue in the app's 4.2.0 release (https://github.com/Automattic/wp-desktop/issues/651) but resolved in the 4.3.0 release. That was a bug coming from Calypso that was fixed in https://github.com/Automattic/wp-calypso/pull/35005, so if your branch has the latest Calypso perhaps there are some cases where the fix doesn't work — although not sure if that's a regression or if there are some cases that weren't tested with that fix.

loremattei:
Thank you @nsakaimbo! 
I smoke tested the app on MacOS and have some good news and some issues I'd like to discuss with you.
**Good:**
1. The app feels a lot faster, at least on my system. 

2. As @rachelmr noted, the back button issue was fixed in Calypso. I pulled the latest Calypso into your branch (test branch here: https://github.com/Automattic/wp-desktop/tree/update/electron-by-nicholas-test) and it works fine also with Electron 4.

**Issues:**
1. The updater seems broken. The app version in your branch is `4.2.0` and, since `4.3.0` has been released, I expect the app to prompt for the update every time it starts, but I haven't seen the prompt so far. I tried to delete all the updater-related stuff in the `Application Support` folder, but it didn't solve. The prompts is shown normally if I change the version to `4.2.0` in the develop branch, so the issue seems to be related to Electron 4. 
You probably already noted this, but I just want to make sure you are aware that the updater class is extended [here](https://github.com/Automattic/wp-desktop/tree/update/electron-by-nicholas-test/desktop/app-handlers/updater), in case it was relevant to the issue. 

2. As you already noted, the spell-checker is broken. With the last Calypso, it looks like the block editor is created for every post, even old-style ones, and that the `classic-editor` block is used for the posts that haven't been updated yet. @rachelmcr is this right, or there's still a way to invoke the classic editor in Calypso?
If this is the case, I think we need to involve Calypso developers to figure out a solution for this issue, that's probably not even completely related to the update. 

3. The build on `CircleCI` is broken. The error in the log is `macOS High Sierra 10.13.6 is required to sign`.
I don't know if you're familiar with CircleCI, anyway in their MacOS containers, the OS version is tied to the XCode version. Per [this post](https://discuss.circleci.com/t/xcode-10-1-image-released/26350), I think that upgrading to a XCode 10.1 container should solve the issue.

loremattei:
I tested it also on Linux and things look a bit better there:
- The build on CircleCi is successful.
- The updater shows the prompt.

nsakaimbo:
> That's curious. This was reported as an issue in the app's 4.2.0 release (#651) but resolved in the 4.3.0 release. That was a bug coming from Calypso that was fixed in [Automattic/wp-calypso#35005](https://github.com/Automattic/wp-calypso/pull/35005), so if your branch has the latest Calypso perhaps there are some cases where the fix doesn't work — although not sure if that's a regression or if there are some cases that weren't tested with that fix.

@rachelmcr You're right - @loremattei was able to confirm that this was in fact addressed. My error - apparently I wasn't testing with the latest-and-greatest Calypso, but glad to hear this is a non-issue. 🎉

rachelmcr:
> With the last Calypso, it looks like the block editor is created for every post, even old-style ones, and that the classic-editor block is used for the posts that haven't been updated yet.  @rachelmcr is this right, or there's still a way to invoke the classic editor in Calypso?

@loremattei This is a per-user, per-site setting, so if you are opted in to the block editor on any given site, the block editor will be used for every post on that site. You can opt out (so the Calypso classic editor is used for every post) by switching to the classic editor at the bottom of the ellipsis menu in the top right.

loremattei:
Thank you @rachelmcr!

@nsakaimbo I've merged your last changes in the test branch and now the build works 👍
Another good news: the prompt for updating is shown in the CI, so I wonder if the problem I saw yesterday is related to my local setup. I'll investigate a bit more. 

A new issue: the tests on the CI are failing because of Calypso not starting. With your branch, I can replicate this also on my local environment, so this should make things easier to debug.

Here is the video of the failing tests on CircleCI:
[e2e-test-run-2019-09-03T07-33-37.mpg.zip](https://github.com/Automattic/wp-desktop/files/3569128/e2e-test-run-2019-09-03T07-33-37.mpg.zip)

nsakaimbo:
Quick update on the issues previously noted, @loremattei:

> 1. The updater seems broken. 

Based on more testing, seems the updater was a non-issue and is in fact working?

> 2. As you already noted, the spell-checker is broken ...  If this is the case, I think we need to involve Calypso developers to figure out a solution for this issue, that's probably not even completely related to the update

Yep - I believe fixing the spellchecker might be a little more involved and deserving of its own project. Based on my research, there are known issues with the spellchecker and iframe content. In addition, newer versions of Electron might have stricter safeguards around iframes.

> 3. The build on `CircleCI` is broken. The error in the log is `macOS High Sierra 10.13.6 is required to sign`.

Updated the macOS version in de1e35a.

dmsnell:
When I've been aware of changes in the past that broke the spell-checker it wasn't a small issue. @nsakaimbo how much effort do you think it would take to fix the spell-checker ahead of this update?

That is, I think our customers will be far more disappointed if we remove that than they will be happy that we upgraded. Thankfully this update looks simple enough like we have some flexibility on when to upgrade it.

Alternatively, are there any updates from this PR that you think we can apply without upgrading Electron so that the actual changeover could take place with a smaller set of changes?

nsakaimbo:
@dmsnell We certainly have some flexibility w.r.t. when the upgrade is deployed. I am aware that upgrading is a high priority at the moment due to some security concerns on Windows, as well as to support new code signing requirements on Mac.

> ... are there any updates from this PR that you think we can apply without upgrading Electron ...?

This PR is fairly limited to changes required by the upgrade. However, I believe 6c2e66c fixes a bug which has actually been a race condition in the current build system, only just made more apparent by the upgrade. We would need to test this out in CI to verify. That commit also adds much-needed logging to the CI test process. 

In addition, #668 is a quality-of-life enhancement that can be merged outside of the upgrade.

> ... how much effort do you think it would take to fix the spell-checker ahead of this update?

This (and associated) PRs are targeted at a feature branch currently, and I agree we shouldn't deploy the upgrade with the regression in the spellchecker. It's hard to estimate how long a fix might take (because ... estimating software). Assuming we get assistance from Calypso to look into this, maybe an additional 1-2 weeks of effort, give or take?

codebykat:
I smoke-tested this and didn't note any issues, and the code looks sensible. I second @dmsnell that breaking spellcheck is going to be a huge pain point from a user perspective and it would be better if we could work on that before merging this in.

loremattei:
@dmsnell @codebykat thanks for looking at this!
I totally agree that we should fix the spellchecker issues before merging this upgrade into `develop`. 

As @nsakaimbo mentioned, we need to upgrade in order to solve some security issues and to be able to leverage some changes in Electron-builder that enable notarization for the MacOS binary. We don't know yet when notarization will become mandatory on MacOS, but Apple mentioned that it would be in _near future_.

@dmsnell @codebykat, if you agree, I'll merge this PR into the feature branch in our repository, as this would make easier to fix the CI configuration (the CI doesn't work on forked PR). Is it ok for you?

dmsnell:
@loremattei it seems unlikely that this update will have any impact on the security issue so I'd be reluctant to prioritize it based on that claim, but if it's in the hot path for fixing the CI issue then that's fine. what's the CI issue?

as to the feature branch I'm new to this flow and having trouble keeping the complexity in my head. if we're not delivering this change to our customers what does rushing it into a branch give us? Please pardon my naïveté.

loremattei:
@dmsnell we have this report: https://github.com/Automattic/wp-desktop/network/alert/package-lock.json/electron/open that will be surely solved by this update.
Also, I _think_ that at least some of the others that GitHub reports are related to Electron dependencies, as there is no report on Calypso and almost every other dependency in WPDesktop is related to Electron.

> what's the CI issue?

We have some automation on CircleCI to build and release the app. Currently, we use pre-built packages for the spellchecker, so we are able to build MacOS, Linux and Windows version on CircleCI, even if it doesn't support Windows build because we just package things. 
With this update, we need to build Windows native binaries, so we need to move Windows builds on AppVeyor or another service, as we currently do on Simplenote.

> as to the feature branch I'm new to this flow and having trouble keeping the complexity in my head. if we're not delivering this change to our customers what does rushing it into a branch give us? Please pardon my naïveté.

This PR is from a forked repository, so we can't test release builds on CI on it, because we don't allow CircleCI to run on forked PRs for security reason (it could leak some secrets). Merging the changes in a feature branch in our repository will enable CI.

I hope this is clearer now, but please feel free to ask any other questions you have!

dmsnell:
> we have this report

Thanks @loremattei - that's a different issue than the one I was given as the reason motivating the change. That linked CVE makes more sense.

> build and release the app

Cool! So does that mean we have a development build somewhere to test?

loremattei:
> Cool! So does that mean we have a development build somewhere to test?
So... they are usually available on CircleCI, but it only happens for not forked PRs. This is forked PR, so CircleCI doesn't do its magic :-) 
This is one reason why merging into a feature branch would help.

PR #669, opened 2019-09-13
[Electron Upgrade] [WIP] Add AppVeyor Config for Windows

nsakaimbo:
### Description:
This PR adds an AppVeyor config file for Windows.

### Motivation and Context:

Since upgrading Electron necessitates we move to native builds on all platforms, we're using AppVeyor to build the application for Windows.

### How Has This Been Tested:
This script has *not* been tested.

### Todos:

- [ ] Replace proper values for `CSC_ENC_KEY`, `CSC_KEY_PASSWORD`, `GH_TOKEN` and `APPVEYOR_RDP_PASSWORD`
- [ ] Verify and test the script, tweak and update as necessary

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,50 @@
+environment:
+  CSC_ENC_KEY: 
+    secure: <TODO>
+  CSC_KEY_PASSWORD:
+    secure: <TODO>
+  GH_TOKEN:
+    secure: <TODO>
+  APPVEYOR_RDP_PASSWORD:
+    secure: <TODO>
+  CSC_LINK: .\resource\certificates\win.p12
+
+image: Visual Studio 2017
+
+configuration:
+  - Release
+
+install:
+  - ps: Install-Product node 10
+  - cinst make
+  - npm ci
+
+for:
+  # default build / .exe / signed
+  - matrix:
+      only:
+        - configuration: Release
+    before_build: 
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -0,0 +1,50 @@
...
+  - matrix:
+      only:
+        - configuration: Release
+    before_build: 
+      - openssl aes-256-cbc -d -in .\resource\certificates\win.p12.enc -out %CSC_LINK% -k "%CSC_ENC_KEY%"
+      - openssl aes-256-cbc -d -in .\resource\secrets\config.json.enc -out .\config.json -k "%CSC_ENC_KEY%"
+    build_script:
+      - make build
+      - make test
+      - 7z x -a -ttar -so release\win-unpacked-x64.tar release\win-unpacked | 7z a -si release\win-unpacked-x64.tar.gz
```
nsakaimbo: Not sure if I'm on the right track here, but the goal was to closely match the current steps in Circle for Windows ([link](https://github.com/Automattic/wp-desktop/blob/develop/.circleci/config.yml#L306)).

We need to use 7-zip in order to create a tar.gz file on Windows, which [comes bundled with the AppVeyor image](https://www.appveyor.com/docs/windows-images-software/#tools). Unlike the `tar` utility for linux, you can't create a `tar.gz` file in one shot with 7-zip, so I cribbed this technique from this [Stack Exchange post](https://superuser.com/questions/339931/how-do-i-use-7-zip-cli-for-windows-to-create-tar-gz-archives). Hopefully it will work, but might need some tweaking if not.

PR #671, opened 2019-09-18
[Upgrade Electron] Appveyor Config

nsakaimbo:
### Description:
This PR adds an AppVeyor config file for Windows.

### Motivation and Context:

Since upgrading Electron necessitates we move to native builds on all platforms, we're using AppVeyor to build the application for Windows.

### How Has This Been Tested:
With minimal syntax changes, I've been able to invoke the makefile in a Windows environment (with MSYS and MINGW, which are both included in AppVeyor's Windows 10 image).

### Blocker

Testing locally reveals there is an open issue against Calypso that prevents the application from being built on Windows 10.

While a fix is pending, one option might be to use a combined Linux/Windows pipeline.

Background/info in [this gist](https://gist.github.com/nsakaimbo/3e783dceddbf5ee0033e46c897a4625d).

### Todos:

- [ ] Given current blocker on Calypso, determine next steps (refer to proposals in the gist linked above)
- [ ] Replace proper values for `CSC_ENC_KEY`, `CSC_KEY_PASSWORD`, `GH_TOKEN` and `APPVEYOR_RDP_PASSWORD`
- [ ] Implement linux/Windows pipeline?
- [ ] Verify and test the script, tweak and update as necessary

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,50 @@
+environment:
+  CSC_ENC_KEY: 
+    secure: <TODO>
+  CSC_KEY_PASSWORD:
+    secure: <TODO>
+  GH_TOKEN:
+    secure: <TODO>
+  APPVEYOR_RDP_PASSWORD:
+    secure: <TODO>
+  CSC_LINK: .\resource\certificates\win.p12
+
+image: Visual Studio 2017
+
+configuration:
+  - Release
+
+install:
+  - ps: Install-Product node 10
+  - cinst make
+  - npm ci
+
+for:
+  # default build / .exe / signed
+  - matrix:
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -0,0 +1,50 @@
...
+  CSC_LINK: .\resource\certificates\win.p12
+
+image: Visual Studio 2017
+
+configuration:
+  - Release
+
+install:
+  - ps: Install-Product node 10
+  - cinst make
```
nsakaimbo: Do we need to explicitly install MSYS/MINGW and Python, etc here as well? Unclear (currently unable to test this myself without access).

PR #672, opened 2019-09-24
Build: Fix SVG errors

### LINKED ISSUE #670: Fix file-loader builds
https://github.com/Automattic/wp-calypso/pull/36139 fixed the server build for Calypso for `file-loader` handled imports from `node_modules`. The same fix will likely be necessary for the Desktop build.

sirreal:
Fixes #670


webpack.shared.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
 				loader: 'babel-loader',
 				exclude: /node_modules/,
 			},
+			FileConfig.loader(),
```
sirreal: We should provide some `file-loader` config here so that the images are put into a public directory and the server can find them.

I've been unable to get that working correctly. They always seem to be written to `build/images`.

sirreal:
Another PR that may help inform this work: https://github.com/Automattic/wp-calypso/pull/36204

PR #673, opened 2019-09-26
Load SVG images (like Gridicons) with webpack file-loader

jsnajdr:
Use the loader to generate correct URLs for server-side rendered images and icons, e.g., in `EnvironmentBadge`. Emitting the files is not necessary, as the actual assets are provided by the client Calypso bundle.

The desktop server bundle is every used only in production mode. Development server runs the classic Calypso as a server running on `calypso.localhost:3000` and points the Electron browser to that server.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6043,6 +6043,46 @@
         "object-assign": "^4.0.1"
       }
     },
+    "file-loader": {
+      "version": "4.2.0",
+      "resolved": "https://registry.npmjs.org/file-loader/-/file-loader-4.2.0.tgz",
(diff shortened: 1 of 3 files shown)


loremattei:
Thank you for looking at this @jsnajdr!
I checked the CI builds and they work as expected. 
The development build, though, seems to be broken. 

If I run `make dev-server` and `make dev` on my local machine, the style looks broken:

Screenshot 2019-09-27 at 15 42 11

loremattei:
I'm merging this in order to fix the CI test builds. We'll address the development build in a different PR.
cc/ @jsnajdr

PR #678, opened 2019-10-16
Macos notarization

loremattei:
This PR adds notarization to the Mac build of WPDesktop.
In order to be notarized, the binary must be built with the hardened runtime flag. The Desktop app seems to be ok with it when the proper entitlements are provided.

Support for hardened runtime has been added on electron-builder 20.41.0, but it requires `electron-updater`  4.0.0 which is not compatible with Electron `1.7`. 
We are working on upgrading Electron on another PR, so I set up a temp stop-gap solution by patching electron-builder on the CI system after npm install. The code I'm using in the patch is the same I sent upstream (electron-userland/electron-builder@7d5f952), so the configuration won't break when we manage to update electron-build.
I don't like this solution too much, so if you have any other suggestion, I'm happy to hear it and to change this PR.

The notarization process runs on tags.

### Test
- This is a build on a commit: https://circleci.com/gh/Automattic/wp-desktop/46957. Check in the logs that the notarization has been skipped.
- This is a build on a tag: https://circleci.com/gh/Automattic/wp-desktop/46964. Check that the notarization has been performed and is successful.
- Verify that there's no sensible data lacked in the logs.
- Download the Mac artifacts and verify that the app still works as expected.

### Release
The PR targets the release branch because Catalina has been released and, while Apple states that notarization won't be required until January 2020, there are reports of additional warnings already popping out that may be confusing for the users.

These changes do not require release notes.

### DIFF
--- a/resource/macos/scheme-patch.diff
+++ b/resource/macos/scheme-patch.diff
@@ -1,0 +1,14 @@
+--- ./node_modules/app-builder-lib/scheme.json	1985-10-26 09:15:00.000000000 +0100
++++ /Users/lore/Desktop/scheme.json	2019-10-14 12:54:44.000000000 +0200
+@@ -1819,6 +1819,11 @@
+                     "description": "Whether to infer update channel from application version pre-release components. e.g. if version `0.12.1-alpha.1`, channel will be set to `alpha`. Otherwise to `latest`.",
+                     "type": "boolean"
+                 },
++                "hardenedRuntime": {
++                    "default": false,
++                    "description": "Whether your app has to be signed with hardened runtime.",
++                    "type": "boolean"
++                },
+                 "electronLanguages": {
+                     "anyOf": [
+                         {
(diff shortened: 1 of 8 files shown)


resource/macos/scheme-patch.diff
```diff
@@ -0,0 +1,14 @@
+--- ./node_modules/app-builder-lib/scheme.json	1985-10-26 09:15:00.000000000 +0100
++++ /Users/lore/Desktop/scheme.json	2019-10-14 12:54:44.000000000 +0200
```
belcherj: Why this change? This will only work on your system.

loremattei: Damn! Good catch! Thank you!
It doesn't actually affect the build because we are applying the patches so it only uses the first line, but it's pretty bad to have. 
I've fixed it.

loremattei:
I fixed the diff files. Thanks for catching it.
This should be ready for another look @belcherj.

Issue #686, opened 2019-10-24
Ensure compatibility with Node 12


sirreal:
Node 12 is the current LTS version and wp-calypso will be looking to migrate soon. The desktop may require some changes and upgrades to ensure it can handle the version change.

See https://github.com/Automattic/wp-calypso/pull/37031#issuecomment-545987889

In tests there, electron-spellchecker didn't seem to build properly.

nsakaimbo:
This issue has been addressed - wp-desktop is now building with latest Calypso on Node 12.

PR #688, opened 2019-10-25
Lock to node 10

sirreal:
This PR does 2 things:
- Replace wp-calypso `.nvmrc` symlink with an `.nvmrc` file referencing the current wp-calypso Node v10 version.
- Make the Node desktop/calypso version mismatch error a warning (don't fail the build)

This should allow to for wp-desktop to run on node v10 while wp-calypso moves ahead.

When wp-desktop is ready to move to Node v12 (and #686 is fixed), this PR can simply be reverted.

This should allow wp-calypso to move ahead with https://github.com/Automattic/wp-calypso/pull/37031

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-calypso/.nvmrc
+10.16.3
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ CURRENT_NODE_VERSION := $(shell node --version | sed -n 's/v\{0,1\}\(.*\)/\1/p')
 # Check that the current node & npm versions are the versions Calypso expects to ensure it is built safely.
 check-node-version-parity:
 ifneq ("$(CALYPSO_NODE_VERSION)", "$(CURRENT_NODE_VERSION)")
-	$(error Please ensure that wp-desktop is using NodeJS $(CALYPSO_NODE_VERSION) to match wp-calypso before continuing. 	Current NodeJS version: $(CURRENT_NODE_VERSION))
+	$(warning Please ensure that wp-desktop is using NodeJS $(CALYPSO_NODE_VERSION) to match wp-calypso before continuing. 	Current NodeJS version: $(CURRENT_NODE_VERSION))
 else 
 	@echo $(GREEN)$(CHECKMARK) Current NodeJS version is on par with Calypso \($(CALYPSO_NODE_VERSION)\) $(RESET)
 endif


sirreal:
This branch is passing with Node 12: https://circleci.com/gh/Automattic/wp-desktop/47765

sirreal:
I'd appreciate if folks could prioritize this review, it's the only known blocker for https://github.com/Automattic/wp-calypso/pull/37031.

cc: @nsakaimbo @loremattei

nsakaimbo:
> This should allow to for wp-desktop to run on node v10 while wp-calypso moves ahead.

To clarify, wp-desktop is currently pinned to Node v7.9.0 (Electron v1.7.x), not Node v10. Unless I'm missing something about the build process or the proposed changes? Ref: [Electron and Node Compatibility](https://github.com/electron/node)

> Make the Node desktop/calypso version mismatch error a warning (don't fail the build)

Is the goal here to build Calypso using Node v12 as part of the wp-desktop build process, while maintaining the Electron v1.7.x (and effectively Node v7.9.0 at runtime)? 

AFAIK, once built, the Calypso server is booted directly from within the Electron application. Electron v1.7.x is actually running a fork of Node v7.9.0, so as long as the Calypso server doesn't expect Node v12 at _runtime_, this might be reasonable. I'm not sure about the implications of the potential mismatch, though - this is still somewhat of a gray area for me. (Worth noting that the runtime version of Node within wp-desktop has been pinned to v7.9.0 while Calypso has been compiled on on Node v10.x for a while now - perhaps this may not be a concern.) If we get this to build we would should QA/test thoroughly.

Side note: I'm not sure what the branching protocol is, but should we merge/test these changes in a feature branch before merging directly into develop?

sirreal:
> To clarify, wp-desktop is currently pinned to Node v7.9.0 (Electron v1.7.x), not Node v10.

Thanks, I did not know that.

> Is the goal here to build Calypso using Node v12 as part of the wp-desktop build process, while maintaining the Electron v1.7.x (and effectively Node v7.9.0 at runtime)?

The goal is to allow Calypso to move to that latest Node LTS (v12), from the previous LTS (v10) it uses now. In order to allow Calypso to move to v12, we're concerned with the build environment for wp-desktop which seems to have issues building on v12. It sounds like there is already a build/runtime version mismatch (v10 vs. electron/v7) and that would continue to exist until wp-desktop updates.

This change would allow wp-desktop to remain on Node v10 at build time while Calypso moves ahead to Node v12 for its purposes.

This highlights some existing problems in wp-desktop, but I don't think they're reasons not to merge this PR.

PR #694, opened 2019-11-04
Fix/update prettier

loremattei:
### Description:
This PR updates `package.json` to install the wp-prettier fork as NPM alias rather than tarball download, following what was done on https://github.com/Automattic/wp-calypso/pull/36259.
It also updates Calypso to the latest.

Thanks @jsnajdr for your help with figuring out this!

cc/ @nsakaimbo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9534,8 +9534,9 @@
       "dev": true
     },
     "prettier": {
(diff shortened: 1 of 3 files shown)


loremattei:
Thanks for looking @jsnajdr!

PR #702, opened 2019-11-12
CI: Use package-lock over npm-shrinkwrap for caches

sirreal:
Necessary after calypso PR:
https://github.com/Automattic/wp-calypso/pull/37493

You can see CircleCI failing to find caches on newer PRs: https://circleci.com/gh/Automattic/wp-desktop/49023

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,13 +43,13 @@ references:
     restore_cache:
       name: Restore npm cache
       keys:
-        - v2-npm-{{ arch }}-{{ checksum "calypso/.nvmrc" }}-{{ checksum "package-lock.json" }}-{{ checksum "calypso/npm-shrinkwrap.json" }}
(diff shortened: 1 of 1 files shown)


sirreal:
🤔 Can we pull in a recent wp-calypso sha?

There's no way (at least not any good way I'm aware of) to get this to work well for wp-calypso tests and tests in this repository while they're out of sync.

loremattei:
IMHO, yes :-) We usually pull on code freezes, but I think that the more frequently we do it, the better because the gap is smaller and we have more chances to catch issues in a timely manner.

sirreal:
I've updated to a recent wp-calypso sha but the tests aren't passing. It seems to hit a timeout after 30s and it's not clear whether it's working or stuck.

I'd appreciate if someone more familiar with the desktop app could help get this landed.

loremattei (approved):
LGTM! Thanks for this update @sirreal!

loremattei:
@sirreal this is a flaky test and I'm pretty sure it's not related to this PR. 
I think @Stojdza is already looking into this.

PR #705, opened 2019-11-13
[TEST] Use Own Prebuilt Spellchecker

nsakaimbo:
### Description:

This PR points the `electron-spellchecker` prebuilt dependency to an updated fork. 

In order to validate the approach, I re-created `v1.1.2` of the `electron-spellchecker` lib, generating binaries compiled against the current Electron version `v1.7.16` for Mac, Windows and Linux. If the application builds and works as expected (on all platforms), then the approach is sound and we could consider using prebuilds as a stop-gap for the Electron upgrade (see Motivation section for more context).

Prebuilding the spellchecker requires forking and updating a total of _four_ repos:

1. [`electron-spellchecker`](https://github.com/nsakaimbo/electron-spellchecker-prebuilt)
2. [`node-cld`](https://github.com/nsakaimbo/cld-prebuilt)
3. [`node spellchecker`](https://github.com/nsakaimbo/spellchecker-prebuilt)
4. [`atom/keyboard-layout`](https://github.com/nsakaimbo/keyboard-layout-prebuilt)

Prebuilt binaries were generated using native Windows, Mac and Linux (Docker) environments for both the current (v1.7.16) and targeted (v4.2.9) versions of Electron.

### Motivation and Context:

**TL;DR:** The fork of the prebuilt repo has not been updated since 2017. In order to upgrade Electron, we may have to take it upon ourselves to update the prebuilt binaries to work around the current inability to build Calypso on Windows 10. (Ideally, we can patch Calypso, but that effort is still up in the air.) 

### How Has This Been Tested:

I smoke-tested the existing application with my own prebuilt binaries, and the behavior on Mac seems identical. (AFAIK the dependency doesn't work - and hasn't worked - on Windows prior)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2511,6 +2511,25 @@
         "to-fast-properties": "^2.0.0"
       }
     },
+    "@paulcbetts/cld": {
+      "version": "github:nsakaimbo/cld-prebuilt#89cb2729b7b394a891e685cee128cca7fc15f594",
+      "from": "github:nsakaimbo/cld-prebuilt",
+      "requires": {
+        "glob": "^5.0.10",
+        "nan": "^2.0.5",
+        "prebuild-install": "^2.4.1",
+        "rimraf": "^2.4.0",
+        "underscore": "^1.6.0"
+      }
+    },
+    "@paulcbetts/spellchecker": {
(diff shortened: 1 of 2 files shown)


nsakaimbo:
Closing - this discovery served its purpose (i.e. learned how spellchecker prebuilds were generated) and is no longer needed.

PR #706, opened 2019-11-15
[build] Fix Build on MacOS 10.15

nsakaimbo:
### Description:

This PR updates `electron-builder` so the application can be built on MacOS 10.15. The updated version of `electron-builder` requires an update to `electron-updater` as well.

### Motivation and Context:

Building on Catalina was failing with the error `Error: Exit code: 2 // ... // Can't locate Mac/Memory.pm in @INC`.

This [issue](https://github.com/electron-userland/electron-builder/issues/3990) was raised in the `electron-builder` repo and addressed as of version 21.2.0.

### How Has This Been Tested:

Run `npm install` and `make build`, per business-as-usual.


nsakaimbo:
@loremattei It looks like version 21.2.0 of `electron-builder` includes [your patch](https://github.com/electron-userland/electron-builder/commit/7d5f952b12406683fd77a5eaac45e8d6b0b9f257#diff-5b5bba5c27e443035436ea831930096b). I confirmed this by running `git tag --contains 7d5f952b12406683fd77a5eaac45e8d6b0b9f257` within the `electron-builder` repo.

I removed the patch files because we shouldn't need them any longer? In which case the CI config should be updated. However I could be mistaken - would appreciate some help double-checking.

loremattei:
@nsakaimbo sounds good! It looks like there's an issue with the  lock file though. Can you take a look?

nsakaimbo:
**Update**: @loremattei This PR is good-to-go, but CI is currently failing pending the fix for the runtime error in #708.

loremattei:
Hey @belcherj @nsakaimbo I smoke tested the app and it seems that this upgrade breaks `electron-updater`. I don't get the upload prompt and I see this error in the console: 
`Checking for update
Error: Error: ENOENT: no such file or directory, open '/Users/lore/Downloads/WordPress.com.app/Contents/Resources/app/dev-app-update.yml'`

So, it looks like `electron-updater` doesn't get the right configuration.

I already saw this in the past with every version > `3.1.2`. I don't know if it depends on Electron version, or if upgrading other packages may fix it.

nsakaimbo:
> ... it seems that this upgrade breaks electron-updater

@loremattei Thanks for the heads up! Will take a look.

PR #708, opened 2019-11-15
Fix Runtime Error: Revert Calypso Update

nsakaimbo:
This PR reverts the Calypso update from #702 to fix a runtime error in wp-desktop.

<p align="center">
  
</p>


nsakaimbo:
> Is there a plan to address the underlying issue? Do we understand the problem?

Agree 100%. I don't understand the inner workings of Calypso enough to further diagnose this issue myself, but I'd certainly be more than happy to pair on this with a Calypso dev. I have limited bandwidth at the moment as I'm looking at upgrading Electron, however I do understand that keeping Calypso up-to-date is a high priority as well.

PR #712, opened 2019-11-20
Electron Upgrade: Build WP-Desktop On Windows

nsakaimbo:
In order to upgrade Electron and transition from the use of prebuilds, we need to build the application natively on all platforms. These changes allow `wp-desktop` to be built on Windows, both locally and on CI (using AppVeyor). Because [Calypso is currently broken and unsupported on Windows](https://github.com/Automattic/wp-calypso/issues/27400), we leverage Docker as a means to build Calypso.

Requirements:
- [Docker](https://docs.docker.com/docker-for-windows/)
- [MinGW](http://www.mingw.org/) toolchain

All of the above tools are included in the images provided by AppVeyor for Windows ([reference](https://www.appveyor.com/docs/windows-images-software/)).

Notes: Docker container RAM has to be set to 8GB as this is a [requirement of Calypso](https://github.com/Automattic/wp-calypso/blob/1f4d67177660485be0b00f81793adf12c0a24189/package.json#L249) - for some reason this requirement is much stricter on Windows Docker than it is on other platforms.

### TODOs

- [x] Amend Makefile
- [x] AppVeyor config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ resource/secrets
 build
 dev-app-update.yml
 test/logs/
+calypso-hash
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "wp-calypso"]
 	path = calypso
-	url = git@github.com:Automattic/wp-calypso.git
+	url = https://github.com/Automattic/wp-calypso.git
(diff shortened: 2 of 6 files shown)


nsakaimbo:
Re-opening to give CI a kick.

PR #718, opened 2019-11-27
Fix Electron-Updater (Broken with Catalina Patch)

nsakaimbo:
**Update 11/28**: Code changes so far seem to have fixed the updater on Mac. However the "differential update" (i.e. blockmap) mechanism on Windows isn't working, and the cause seems fairly deep-rooted in the internals of `electron-builder` and `electron-updater` and inconsistencies with the very stale version of Electron currently in use.

In order to get the application to build on Mac OS 10.15, we had to upgrade electron-updater to v21.2.0 in #706. However, because of incompatibilities with Electron 1.7.16, the auto-update mechanism in develop (as of SHA 2c0a992) was broken.

### Description:

The newer version of `electron-updater` expects an updated interface to select Electron APIs (which will not be available until Electron is upgraded). This PR (which should be reverted with the upgrade) patches Electron for the following APIs:

- Add `app.isPackaged` (indicating whether or not the application is packaged for release)
- Modify the internals of new `electron-updater` to use the (outdated) version of of the call to `setFeedURL`

### How Has This Been Tested:

- Change the `version` number in package.json (to an earlier version than current)
- Build and install the application with `make build CONFIG_ENV=release` config
- Note that the auto-updater prompt is displayed. The application should be updated if "Update and Restart" is selected.

### Screenshots:

Screen Shot 2019-11-26 at 3 10 29 PM

### Todos:

- [x] Confirmed working on Mac
- [ ] Confirmed working on Linux
- [ ] ~~Confirmed working on Windows~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,3 +169,8 @@ clean:
 	@rm -rf .$/build
 
 .PHONY: test build-source
+
+apply-patch:
+	@echo "Applying module patches..."
+
+	patch "./node_modules/electron-updater/out/MacUpdater.js" "./resource/macos/electron-updater.patch" 
--- a/desktop-config/config-updater.json
+++ b/desktop-config/config-updater.json
@@ -10,6 +10,7 @@
 		"downloadUrl": "https://github.com/Automattic/wp-desktop/releases",
 		"apiUrl": "https://api.github.com/repos/Automattic/wp-desktop/releases",
 		"delay": 2000,
-		"interval": 600000
+		"interval": 600000,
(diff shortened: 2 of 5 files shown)


nsakaimbo:
Closing  as we're reverting the `electron-updater` upgrade in #718. Fixing the blockmap updater on Windows seems to be fairly deep-rooted and remains unresolved. With the Electron upgrade nearing completion, the time/effort may not be worth it to fix in the short term.

PR #731, opened 2019-12-27
[e2e] Add "Sign up" test to e2e tests suite

Stojdza:
### Description:
This PR adds a new test to e2e tests suite - Sign up test. 

Similar to wp-calypso Sign up specs, this is a basic sign up process with a free domain and free plan chosen. 

### Motivation and Context:

Expand wp-desktop e2e tests suite. Let's start with the sign up flow and evaluate benefits :) 

### How Has This Been Tested:

Run tests locally or in CircleCI. Make sure they are passing. Some inconsistencies still could happen. If that's the case, please share CircleCI URL or the error message from your console in this PR so I can investigate further.


test/tests/e2e.js
```diff
@@ -84,36 +87,96 @@ describe( 'Publish a New Post', function() {
...
+			process.env.E2EPASSWORD
+		);
+	} );
+
+	step( 'Can then see the onboarding checklist', async function() {
+		const checklistPage = await ChecklistPage.Expect( driver );
+		return await checklistPage.isChecklistPresent();
+	} );
+
+	// 	TODO:delete account
```
nsakaimbo: @Stojdza Did you intend to leave TO-DO/comment block?  Perhaps we can make this a GH issue instead, once this is merged?

nsakaimbo:
@Stojdza 

This is excellent - really nice addition to the suite of tests.

Quick question, just to clarify - do we use the new account to run through the existing tests (i.e. create blog post, etc), or is this a standalone test?

Also, minor nit - do you think we might be able to squash some commits in this branch? (I suppose if we use squash-and-merge it won't matter - we currently don't have a consistent PR merging strategy in wp-desktop).

test/tests/e2e.js
Stojdza: I was planning to leave TODOs, but GH issue sounds like a way better option! :) Thanks!

Stojdza:
Thanks, @nsakaimbo! 

> do we use the new account to run through the existing tests (i.e. create blog post, etc), or is this a standalone test?

Existing tests use already existing account, I didn't change that. For Sign up test new, random, email address is generated and used. I used [Mailosaur email inbox](https://github.com/Automattic/wp-desktop/blob/update/e2e/add-signup/test/tests/e2e.js#L124), but I could put anything here since we don't verify email address during the test.

> do you think we might be able to squash some commits in this branch?

Squash and merge, definitely! It would be messy to merge all these commits without squash.

Stojdza:
@nsakaimbo I've merged the latest changes from `develop` branch and now build is failing. Do you know what is causing this and how to solve it?

nsakaimbo:
@Stojdza The build errors are because this branch is using a Calypso submodule version that doesn't exist in the Calypso repo. (In addition, the submodule version in this branch is very stale with respect to latest Calypso and needs to be rebased.)

```
error computing cache key: template: cacheKey:1:88: executing "cacheKey" at <checksum "calypso/package-lock.json">: error calling checksum: open /home/circleci/wp-desktop/calypso/package-lock.json: no such file or directory
```

```
ERROR in ./desktop/server/server.js
Module not found: Error: Can't resolve 'server/boot' in '/home/circleci/wp-desktop/desktop/server'
```

- If Calypso changes are required to support this work, then the submodule revision should point to a Calypso revision that actually exists the Calypso repo. 
- If you need to update Calypso, you need to ensure that whatever changes you're making to Calypso are pushed to a branch in the Calypso repo, and the SHA (`HEAD`) of that branch matches the submodule revision included in wp-desktop. You can see an example of how to do this in #749.
- If you didn't intend to include Calypso changes, the easiest thing to do would be to make sure the submodule revision being used in this branch matches that of `develop`.

Stojdza:
> If you didn't intend to include Calypso changes, the easiest thing to do would be to make sure the submodule revision being used in this branch matches that of `develop`.

I didn't intend to touch Calypso. If I did, it was unintentional. What's the best way to compare (and change) local Calypso revision with the one from `develop` branch? So far I didn't succeed with that.

nsakaimbo:
@stojdza I’m guessing you must have used something like `git add .` after having switched from a branch with an older version of Calyspo. Basically you want to make sure this branch doesn’t have a diff of Calypso with respect to `develop`. There’s probably a few ways you can fix. 

The cleanest way is probably an interactive rebase - edit the commit in which the Calypso diff was added, and remove Calypso from your staged files for that commit. Then force-push your branch to remote after the rebase is complete.

PR #738, opened 2020-01-10
[mac] Enable LSFileQuarantineEnabled for OSX

nsakaimbo:
This change makes it so files created by the application (and subprocesses) are quarantined by default. When quarantining files, the system automatically associates the app name, bundle id with the quarantined file whenever possible. This mitigates attacks via remote code execution.

This change is in response to a recently revealed vulnerability in which a user editing a post may run malicious code by double-clicking an executable downloaded via Opt+Click. This would bypass Mac OS's Gatekeeper completely. 

Adding the `LSFileQuarantineEnabled` flag ensures that Gatekeeper will correctly flag such files and prevent them from running.

### How Has This Been Tested:
Behavior in both the current release of the application (v4.6.0) and a custom build with this patch was tested with a variety of media (`.terminal` executable and media such as `.png` and `.mp3`).

- The current stable release (v4.6.0) allowed the executable to run. A build with this patch did not.
- This patch does not appear to interrupt expected interaction with other media types (such as audio and images). 

<p align="center">
Screen Shot 2020-01-10 at 5 54 10 PM
</p>

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,10 @@
       "category": "public.app-category.social-networking",
       "entitlements": "./resource/macos/entitlements.mac.plist",
       "entitlementsInherit": "./resource/macos/entitlements.mac.inherit.plist",
-      "hardenedRuntime": true
+      "hardenedRuntime": true,
+      "extendInfo": {
+        "LSFileQuarantineEnabled": true
+      }
     },
     "dmg": {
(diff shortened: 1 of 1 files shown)


bummytime:
Beautiful @nsakaimbo! I agree with @dmsnell — seems to be a no brainer to :shipit:

PR #739, opened 2020-01-13
Adapt desktop server build after Calypso server folder rename

jsnajdr:
This will fix the desktop build after https://github.com/Automattic/wp-calypso/pull/38190 is merged. Reacts to rename of the `calypso/server/` folder to `calypso/client/server/`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,7 @@ references:
     - resource/certificates/win.p12
     - resource/certificates/mac.p12
     - calypso/public
-    - calypso/server
+    - calypso/client/server
     - calypso/config
     - calypso/package.json
     - calypso/packages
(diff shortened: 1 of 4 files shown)


jsnajdr:
@nsakaimbo After https://github.com/Automattic/wp-calypso/pull/38190 got merged, this PR adapts the desktop app to start working with the new Calypso folder structure again.

Please have a look. What needs to be done is to update the `calypso` submodule to latest Calypso `master` (I didn't do it in this PR, only locally) and verify that the app compiles and works.

jsnajdr:
Hi @nsakaimbo :wave: I pushed the `calypso` submodule update, too. Let's see what CI says about it. I was not sure what's your policy for such updates.

nsakaimbo:
@jsnajdr Thanks for including the submodule update! Generally speaking it's probably cleaner to merge submodule updates separately, but where a change to wp-desktop depends on new changes to wp-calypso, probably makes sense to include the submodule update so we can verify in CI.

PR #742, opened 2020-01-15
[.gitignore] Ignore Decrypted Certs

nsakaimbo:
### Description

This PR ensures the decrypted Mac and Win code signing certificates are ignored so they are never committed to the repo by mistake (easy to do with `git add .`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ publish/node_modules/
 resource/secrets
 build
 dev-app-update.yml
+resource/certificates/mac.p12
+resource/certificates/win.p12


.gitignore
```diff
@@ -8,3 +8,5 @@ publish/node_modules/
 resource/secrets
 build
 dev-app-update.yml
+resource/certificates/mac.p12
+resource/certificates/win.p12
```
dmsnell: should we blacklist this whole directory?

```gitignore
resource/certificates
```

nsakaimbo: @dmsnell 

>should we blacklist this whole directory?

Good question. We could blacklist the entire directory, except the unencrypted assets are in there as well ([link](https://github.com/Automattic/wp-desktop/tree/develop/resource/certificates)). If we ever need to update them we should probably exclude things piecemeal. We certainly could re-factor into a different folder structure, but that change would be a little more substantial as our Circle config uses the existing locations as well.

dmsnell: no need to refactor, just double-checking that it's not more convenient to collapse the rule into one

PR #749, opened 2020-01-21
Electron upgrade v4.2.9 fix spellchecker

nsakaimbo:
### Description

This PR restores spellchecker functionality, which was lost as a result of upgrading Electron in wp-desktop. The expected behavior is fixed by:

#### 1. Attaching the spellchecker after the editor iframe is loaded.

In order to address a limitation of updated Electron's spellcheck API, we need to ensure that the spellchecker is attached after the Gutenberg editor is fully loaded. We could have listened for the window event [directly](https://github.com/Automattic/wp-calypso/blob/e1face4840c4d7710a66d76d90bf1ff810dede8b/apps/wpcom-block-editor/src/calypso/features/iframe-bridge-server.js#L798), but this wouldn't be aligned with the existing architecture that relays Calypso events to Electron via Calypso's `client/lib/desktop` module. Instead, we create a corresponding redux event in Calypso and send _that_ to Electron's preload script.

See the corresponding Calypso PR at github/Automattic/wp-calypso#38986.

#### 2. Circumventing the spellchecker's early exit condition

`electron-spellchecker`'s currently implementation assumes that content in the document body is being actively edited when `attachToInput` is called.

https://github.com/electron-userland/electron-spellchecker/blob/7227d61415c47dfa247828da3b62bbb3330095aa/src/spell-check-handler.js#L206

The Calypso editor doesn't satisfy this assumption when a post (either existing or new) is loaded. As a workaround, we inject a dummy observable to prevent the method from exiting before the user begins to make edits.

### To Test

1. Build and run the application, ensuring to check out the appropriate patch of Calypso that wires up the `editor-loaded` event
2. Start editing an existing or new post
3. Misspellings should be highlighted in active text areas, per the screenshots attached

<p = align="center">
Screen Shot 2020-01-16 at 7 19 21 PM
<br>
Screen Shot 2020-01-16 at 7 19 11 PM
</p>

### DIFF
--- a/public_desktop/preload.js
+++ b/public_desktop/preload.js
@@ -1,6 +1,8 @@
-
 const { ipcRenderer, remote } = require( 'electron' );
 const { SpellCheckHandler, ContextMenuListener, ContextMenuBuilder } = require( 'electron-spellchecker' );
+const { Observable } = require( 'rxjs/Observable' );
+require( 'rxjs/add/observable/from' ); // necessary for providing Observable.from() to Observable (see usage below)
+
 const debugModule = require( 'debug' );
 
 const desktop = remote.getGlobal( 'desktop' );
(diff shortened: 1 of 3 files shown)


nsakaimbo:
Note to reviewer: Ignore the `continuous-integration/appveyor/pr` job - not sure why that's showing up here (only 1 appveyor job is supposed to be attached to any given PR - probably something in AppVeyor settings).

public_desktop/preload.js
```diff
@@ -1,37 +1,47 @@
-
 const { ipcRenderer, remote } = require( 'electron' );
 const { SpellCheckHandler, ContextMenuListener, ContextMenuBuilder } = require( 'electron-spellchecker' );
+const { Observable } = require( 'rxjs/Observable' );
+require( 'rxjs/add/observable/from' );
```
dmsnell: it would be best to add an explanatory comment here why we need to include this and why we're including it for its side-effects

nsakaimbo: @dmsnell Thanks for the review!

There's a fairly detailed explanatory note regarding the use of RxJS a few lines down into the file. I think this should be sufficient? However let me know if I can add or clarify any details.

https://github.com/Automattic/wp-desktop/pull/749/files#diff-c3c73cf27ad747e5e2d364fc3bb986b9R30

> I'm not seeing the mechanism for how the dummy$ gets its input events but I assume that is working anyway?

Yep - it's working. 😅Seems adequate to initialize the observable as empty ( using empty array `[ ]` ) to bypass the `null` check in the spellchecker source code.

dmsnell: > There's a fairly detailed explanatory note regarding the use of RxJS a few lines down into the file

we won't suffer by explaining it here either. us the one you're referring to the one in the new line 31? the one that doesn't mention `rxjs` 😉 ?

we don't need a big explanation but coupling the explanation with the line itself will help prevent someone from removing it in the future. in any case, I don't see an explanation anywhere for why we need a side-effect-only import, especially since in the line above we're importing a named object

```js
require('rxjs/add/observable/from'); // necessary for providing Observable.from() to Observable
```

PR #757, opened 2020-01-30
[Electron] Upgrade To v4.2.9

nsakaimbo:
### Description

This PR upgrades wp-desktop from electron to v4.2.9. A number of breaking changes were fixed along the way, and a few enhancements were made as well.

- [x] Upgrade Electron to v4.2.9 (fcb4757)
- [x] Dependencies: `electron-builder`, `electron-updater` (fcb4757)
- [x] Dependencies: replace `electron-spellchecker` prebuilt with source (6a928e0)
- [x] Dependencies: revert manual node module patches required to notarize the app with Electron v1.x (584acab)
- [x] Fix breaking API changes since v1.x: 
    * `crashReporter` (85c2dd8)
    * `makeSingleInstance` (edcb826)
    * Undocumented breaking change: `session.setProxy` (17e5e4e)
-  [x] CI: Fix e2e test breakages (378a006), update `electron-chromedriver` (
927a4b9), update user-agent info (ef6f4c2)
-  ~~[x] CI [Enhancement]: Add e2e utility script to prep for local and cross-platform invocation (0087065)~~ (Reverted - will revisit as part of a dedicated task)
-  [x] [Windows] Makefile fixes to enable native Windows builds (cc67de8)
 - [x] CI [Mac]: Update Xcode image to 10.1.0 (f348550)
-  [x] CI [Linux]: Add `libxext-dev` and `libx11-dev` for spellchecker compilation on Linux (543f679)
-  [x] CI [Windows]: Enable native Windows builds in Circle CI (41d245c)
-  [x] Fix regressions in `electron-spellchecker` behavior (b283fce and Calypso PR github/Automattic/wp-calypso#38986)
- [x] **Bonus**: The spellchecker is not only fixed, but working on Windows and Linux now! 🎉
- [x] **Bonus**:  Due to an incompatibility with Mac OS Catalina, we lost our ability to build DMGs locally (because of the very stale version of `electron-builder`). This upgrade restores the ability to build all application artifacts locally. 🎉 

#### Code Review
 
- With the exception of enabling native Windows builds in Circle CI (e420f9a and a5137ac), the entirety of this code has already been subject to review in upstream PRs, and has been re-organized here for ease-of-navigation.

#### Manual testing:  

- Feel free to give the built application a smoke-test. Artifacts for all platforms built from this branch are located here:

[Build 52836: Artifacts](https://circleci.com/gh/Automattic/wp-desktop/52836#artifacts/containers/0)

Navigate to the "Artifacts" tab to download the .exe, .deb or .dmg for Windows, Linux and Mac, respectively.

<p align="center">
Download Artifacts
</p>

Any and all bugs are worth making note of. However if the bug already exists in the current stable release (v4.7.0 in Github Releases), this will not be considered a regression on existing stable code and will not be in scope for fixes in this branch. (However, please do file an issue if the issue doesn't already exist!)

**Update**:

- [x] The application has been smoke-tested on all platforms and appears to be working as expected.


package.json
```diff
@@ -15,13 +15,13 @@
   "homepage": "https://desktop.wordpress.com/",
   "main": "build/desktop.js",
   "scripts": {
-    "preinstall": "make check-node-version-parity",
-    "postinstall": "make apply-patches && make rebuild-deps",
+    "preinstall": "make check-node-version-parity && make check-electron-version-parity",
```
dmsnell: I saw some other examples that had added these chains (in previous commits) but we don't need to and if we do we can ironically work around what `make` offers.

these changes can go straight into the `Makefile` and rely on `make`'s own dependency-management so that they run in sequence

dmsnell: if anything here I would expect us to create a new `make` target like the following

```make
.PHONY: check-version-parity
check-version-parity: check-node-version-parity check-electron-version-parity
```

nsakaimbo: Thanks for the suggestion - agree this is how it really _should_ be done. Fixed in ed8b982.

nsakaimbo:
**Update**: 

While it works, seems there's room to further optimize how the `e2e` script is invoked (we can probably do this exclusively using `make` - thanks for the suggestion, @dmsnell!). However, the script is currently a value-add at this point and not an essential part of the upgrade. Reverting in e21aac4 to avoid confusion but will revisit down the road.

Tracking in: https://github.com/Automattic/wp-desktop/issues/758

Makefile
```diff
@@ -57,7 +66,7 @@ dev-server: checks
 	@echo "|                                                |"
 	@echo "+------------------------------------------------+$(RESET)\n\n"
 
-	$(MAKE) desktop$/config.json CONFIG_ENV=$(CONFIG_ENV)
+	$(MAKE) desktop/config.json CONFIG_ENV=$(CONFIG_ENV)
```
dmsnell: was this change intentional? the `$/` is a guard for cross-platform file references

Makefile
```diff
@@ -70,26 +79,27 @@ dev: DEBUG = desktop:*
...
+BASE_CONFIG := ./desktop-config/config-base.json
+TARGET_CONFIG := ./desktop-config/config-$(CONFIG_ENV).json
 
-BASE_CONFIG := $(THIS_DIR)/desktop-config/config-base.json
-ENV_CONFIG := $(THIS_DIR)/desktop-config/config-$(CONFIG_ENV).json
-
-.PHONY: desktop$/config.json
-desktop$/config.json:
-ifeq (,$(wildcard $(ENV_CONFIG)))
+.PHONY: desktop/config.json
```
dmsnell: same here - did you experience problems with `$/`?

Makefile
```diff
@@ -98,46 +108,51 @@ calypso-dev:
...
 # Combined steps for building app 
 build: build-source package
 
 # Perform checks
-checks: check-node-version-parity secret
-
+checks: check-version-parity secret
 
 # Check for secret and confirm proper clientid for production release
+SECRETS := ./calypso/config/secrets.json
```
dmsnell: the `CALYPSO_DIR` variable should be an absolute reference and therefore should work regardless of where the `Makefile` was invoked. when we move to relative file references we change the links based on the current working directory from whence this was invoked, so if we do `make wp-desktop/` or `make ..` then we'll probably point to something unexpected

Makefile
```diff
@@ -169,9 +195,3 @@ clean:
 	@rm -rf .$/build
 
 .PHONY: test build-source
-
-apply-patches:
-	@echo "$(CYAN) Applying patches...$(RESET)"
-
-	patch -p1 < ./resource/macos/macPackager-patch.diff
-	patch -p1 < ./resource/macos/scheme-patch.diff
```
dmsnell: There are some repetitive elements here I'd like to consider undoing (unless we have concrete issues triggering the changes).

 - we're removing `$/` in favor of `/` which at least used to prevent breakage in Windows. when running in WSL and some other environments I know this won't be as much of a factor, but if that's the case it's probably better to leave the changes out of this patch and do so entirely in one go in a PR like "Remove no-longer-necessary `$/` substitution"

 - Maybe green looks better than cyan but I'm not seeing a warrant for changing it or a need to do so as part of the Electron upgrade. If we don't like it we could more appropriately address that in its own PR "Change the color of messages to more closely indicate status" or something like that. Here it's conflating the responsibility of what's necessary for the change and what's an unrelated styling issue.

.gitignore
```diff
@@ -10,3 +10,4 @@ build
 dev-app-update.yml
 resource/certificates/mac.p12
 resource/certificates/win.p12
+test/logs/
```
dmsnell: don't know if it matters or not but the trailing slash I'm pretty sure isn't necessary and may be unwanted, differentiating things _in_ `test/logs` vs `test/logs` itself

Makefile
```diff
@@ -12,14 +14,21 @@ THIS_MAKEFILE_PATH := $(word $(words $(MAKEFILE_LIST)),$(MAKEFILE_LIST))
 THIS_DIR := $(shell cd $(dir $(THIS_MAKEFILE_PATH));pwd)
 NPM_BIN_DIR = $(shell npm bin)
 
+NPMRC_ELECTRON_VERSION := $(shell npm config get target)
+PACKAGE_ELECTRON_VERSION := $(shell node -e "const electronVersion = require('./package.json').devDependencies.electron; console.log( electronVersion )")
```
dmsnell: might as well simplify here…

```js
node -e "console.log(require('./package.json').devDependencies.electron)"
```

…that is, there's no value in creating the middleman variable

Makefile
nsakaimbo: Yep - this is very intentional. 

On Windows, it's common for devs to use Git bash. Git bash (derived from MSYS2) expects *nix-style paths, and invoking `make desktop\config.json` _won't_ work (because of how MSYS  attempts to translate file paths). Seems coding it this way was assumed to work, and apparently wasn't tested because it actually doesn't. 

One of the key advantages of using MSYS/Git/Windows bash is, ideally, we should be able to use forward-slashes and `:` path separators and let the environment handle the paths accordingly (there are exceptions which we can mark for exclusion, if/where necessary).

TL;DR I'd rather avoid as much platform-specific exceptions like this, and I've confirmed that this works just fine (i.e. you can invoke `make desktop/config.json` on Windows and it will do the right thing (i.e. execute based on whether or not the path exists).

Makefile
nsakaimbo: Yep - this in fact didn't work on Windows (see comment above).

desktop/lib/app-instance/index.js
```diff
@@ -25,13 +25,11 @@ AppInstance.prototype.anotherInstanceStarted = function() {
 };
 
 AppInstance.prototype.isSingleInstance = function() {
-	let shouldQuit;
+	const gotTheLock = app.requestSingleInstanceLock()
 
-	shouldQuit = app.makeSingleInstance( this.anotherInstanceStarted.bind( this ) );
-
-	if ( shouldQuit ) {
+	if ( !gotTheLock ) {
```
dmsnell: `gotTheLock` might seem a bit awkward. if we don't _need_ this variable and we don't use it beyond determining if this call succeeded, we can consider inlining it.

```js
AppInstance.prototype.isSingleInstance = function() {
  if (app.requestSingleInstanceLock()) {
    return true;
  }

  debug( 'App is already running, quitting' );
  app.quit();
  return false;
}
```

Makefile
nsakaimbo: @dmsnell 

- The use of `$/` actually doesn't work on Windows, so this is a necessary change (see comment above) - feel free to verify/test this locally. IMO we shouldn't have these platform-specific snippets if we don't need them.

- The change to green is an artifact of the fact that until a few days ago, we thought we would be going with AppVeyor for our CI solution on Windows - I found it much easier to distinguish the Green from Cyan in contrast to AppVeyor's choice of background color, which helped make navigating the build logs much easier. That being said, since we're going with Circle I'm happy to revert to the original color and minimize the overall diff.

.gitignore
nsakaimbo: Using `test/logs` might be sufficient, however in this case was attempting to be consistent with the rest of the file: `node_modules/`, `publish/node_modules/`. 

If the trailing slash doesn't make a difference, I'd lean toward being stylistically consistent.

Makefile
nsakaimbo: Update: Reverted to usage of cyan in 7e0742a

Makefile
nsakaimbo: Yeah - good call. For some reason I was having trouble with earlier on Windows, but it's not an issue any longer. Reverted to absolute paths in 979b8bb.

Makefile
nsakaimbo: Good suggestion - refactored in a99c72b

Makefile
dmsnell: > Seems coding it this way was assumed to work, and apparently wasn't tested because it actually doesn't.

that's not entirely true. it was broken, then this was added, then it worked, now it's apparently broken again in yet-another Windows environment 😉 

if we're going to remove it then maybe we should remove it everywhere, and probably in a precursor PR

.gitignore
dmsnell: eh, looks like we're inconsistent in this file already, but `node_modules/` and `publish/node_modules/` are the outliers. use whichever; though the slash is significant

> If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories.
https://git-scm.com/docs/gitignore

I think the `publish/node_modules/` is even obsoleted by `node_modules/` 🤔

Makefile
nsakaimbo: Also worth noting that the removal of `$/` is limited to this target of the Makefile. It's usage may/may not be necessary in other existing targets, but that's TBD and not in scope for the time being.

Makefile
```diff
@@ -70,23 +79,24 @@ dev: DEBUG = desktop:*
...
 endif
-	@node -e "const base = require('$(BASE_CONFIG)'); let env; try { env = require('$(ENV_CONFIG)'); } catch(err) {} console.log( JSON.stringify( Object.assign( base, env ), null, 2 ) )" > $@
+	@node -e "const base = require('$(BASE_CONFIG)'); let env; try { env = require('$(TARGET_CONFIG)'); } catch(err) {} console.log( JSON.stringify( Object.assign( base, env ), null, 2 ) )" > $@
 	
-	@echo "$(GREEN)$(CHECKMARK) Config built $(if $(EXTENDED),(extended: config-$(CONFIG_ENV).json),)$(RESET)"
+	@echo "$(CYAN)$(CHECKMARK) Config built $(if $(EXTENDED),(extended: config-$(CONFIG_ENV).json),)$(RESET)"
 
 # Build calypso bundle
 build-calypso: 
+	@echo "Building calypso..."
```
dmsnell: just stylistically speaking if we don't need colors we can use `$(info …)` and `$(warning …)` - the `@echo` invocation creates a new shell process but the macros/functions output from within the `make` process itself

desktop/lib/app-instance/index.js
nsakaimbo: Refactored in cbd6740

Makefile
dmsnell: oh okay. well, things may break. if they do we'll fix them. carry on

Makefile
nsakaimbo: Good call.  Changed in d3e158f.

Will probably do another pass over the entire Makefile in a separate PR.

Makefile
nsakaimbo: > ... it was broken, then this was added, then it worked, now it's apparently broken again in yet-another Windows environment

Yeah. I'm beginning to learn Windows can be ... funny that way. 🤔 

Interestingly enough this error was consistent in AppVeyor as well, which also uses MSYS2 - that could be the common denominator.

Makefile
```diff
@@ -12,14 +14,21 @@ THIS_MAKEFILE_PATH := $(word $(words $(MAKEFILE_LIST)),$(MAKEFILE_LIST))
...
 RESET = `tput sgr0`
 
 CALYPSO_DIR := $(THIS_DIR)/calypso
 
-CHECKMARK = ✓
+ifeq ($(OS),Windows_NT)
+	CHECKMARK = OK
+else
+	CHECKMARK = ✓
+endif
```
dmsnell: if we only use `OK` we won't need this variable at all and then the output will likely be a little more consistent across platforms

nsakaimbo: Amended in 2a57d82.

belcherj:
Yay!!!!

PR #778, opened 2020-02-19
Add End-to-End Testing Utility

nsakaimbo:
### Description

This PR adds tooling to enable easier invocation of the end-to-end suite locally, and sets the stage for cross-platform invocation in the near future.

To-date, running tests locally has been a clumsy and cumbersome process in which developers had to cross-reference the existing Circle CI script to determine the steps required to invoke the test suite. This involved multiple manual steps:

1. Setting up the necessary environment variables (test credentials)
2. initializing the application 
3. Initializing chromedriver
4. Running the tests via mocha, and finally
5. Shutting down and re-starting chromedriver and the application between test runs

Debugging and augmenting the test suite locally has been significantly hampered by the lack of tooling to enable faster and more efficient iteration. The goal of this utility is to significantly reduce this overhead such that invoking the tests only requires two steps: 

1. Setting up the necessary environment variables (which are now made explicit), and
2. invoking the tests with `make e2e` or `npm run e2e` 

At the end of teach test run (or when the suite errors or is interrupted via `CTRL+C`), all processes are automatically shut down. Automated process cleanup allows repeated tests runs with little to no manual intervention between each test run.

### To Test

1. Set required environment variables (username, password, email)
2. Invoke tests with `make e2e` or `npm run e2e`

### Alternatives

Alternatives using `make` and `bash` were explored, but ultimately did not function as desired. `bash` in particular seems promising and might be worth re-visiting down the road.


test/run/checks.js
```diff
@@ -0,0 +1,28 @@
+
+function usernameExists() {
+    return ( process.env.E2EUSERNAME && process.env.E2EUSERNAME !== '' ) ? true : false;
```
dmsnell: fun fact: this is equivalent to `Boolean( process.env.E2EUSERNAME && … )` and also to `!! (process.env.E2EU…)`

just noting since the ternary is there a bit surprisingly - was there something odd about the condition that prevented us from casting it to a boolean?

anyway, not a big deal

test/run/run-e2e-tests.js
```diff
@@ -0,0 +1,53 @@
...
+    if ( app ) {
+        app.kill();
+    }
+}
+
+async function run() {
+    try {
+        checkCredentials();
+
+        const timestamp = ( new Date() ).toJSON().replace( /:/g, '-' );
```
dmsnell: can we change this to be some explicit format? I find it near impossible to know the intention behind it, especially if for some reason the JSON serialization changes and we end up with formats that we don't expect.

```js
const formatYMD = instant => {
  const year = instant.getFullYear();
  const month = `${ instant.getMonth() + 1 }`.padStart(2, '0');
  const day = `${ instant.getDate() }`.padStart(2, '0');

  return `${ year }-${ month }-${ date }`;
}

…

const timestamp = formatYMD( new Date() );
```

inline isn't so bad: it's still explicit but it is uglier on account of needing to cast numbers to strings

```js
const now = new Date();
const timestamp = `${now.getFullYear()}-${`${now.getMonth() + 1}`.padStart(2,'0')}-${`${now.getDate()}`.padStart(2, '0')}`
```

test/run/init_app.js
```diff
@@ -0,0 +1,20 @@
+const { spawn } = require( 'child_process' );
+const { BUILT_APP_DIR } = require( './paths' );
+
+function initApp( log ) {
+    const startCmd = './'
```
dmsnell: was this more significant before? what's the purpose of splitting it from `./WordPress.com`?

test/run/init_app.js
```diff
@@ -0,0 +1,20 @@
...
+        '--start-maximized', '--remote-debugging-port=9222' ],
+        { stdio: [ 'ignore', log, log ], detached: true, cwd: BUILT_APP_DIR }
+    );
+
+    app.on( 'error', err => {
+       throw new Error( 'failed to initialize app: ', err );
+    } );
+
+    return app;
+}
```
dmsnell: Just playing around here…

```js
function run( cwd, command, args, output ) {
  const app = spawn( command, args, { stdio: [ 'ignore', output, output ], detached: true, cwd } );
  app.on( 'error' => { throw new Error( `failed to initialize ${command}` ); } );

  return app;
}

const app = run( BUILT_APP_DIR, './WordPress.com', [
  '--disable-renderer-backgrounding',
  '--disable-http-cache',
  '--start-maximized',
  '--remote-debugging-port=9222',
], appLog );

await delay(5000);

const driver = run( PROJECT_DIR, 'npx', [
  'chromedriver',
  '--port=9515',
  '--verbose',
], driverLog );
```

There's a part of me feeling like we could make a simple abstraction over running the process and then eliminate these two `init…` files. I'm nervous about having the _straightforward_ logic of starting the tests be obscured by being hidden in a number of split up files.

Even the `appLog` and `driverLog`, for example, could be inlined in `run-e2e-tests.js`

For transparency, my anxiety about this is from the fact that the test runner systems are always one-off projects and have their own language and style. In these cases I definitely value having more stuff in one file where I can see it in my editor all at once over separating the pieces into separate files and modules.

While it may seem to disagree with my play above to introduce the process abstraction I'm still drawing a line between "this is a thing that spawns a process" and "start the app somehow" and "start the driver somehow." Whereas having the startup commands and CLI args in the test runner itself the mechanisms going on are obvious even if there's something hidden in the process abstraction (redirection of `stdout` and `stderr`). The pattern should be obvious enough from the `cmd, [ arg1, arg2 ]` syntax

test/run/run-e2e-tests.js
```diff
@@ -0,0 +1,53 @@
...
+        driver.kill();
+    }
+    if ( app ) {
+        app.kill();
+    }
+}
+
+async function run() {
+    try {
+        checkCredentials();
```
dmsnell: I'l make a similar comment here about my own anxiety and obscuring what's happening vs. hiding _how_ it's happening. this doesn't give me much information about the credentials even though I assume it's checking _something_.

Another formulation might abstract the check and expose the things being checked…

```js
async function run() {
  const requriedENVs = [ 'E2EUSERNAME', 'E2EPASSWORD', 'E2E_MAILOSAUR_INBOX' ];
  const missingENVs = requiredENVs.filter( name => ! process.env[name] || process.env[name] === '' );
  if ( missingENVs.length ) {
    throw new Error( `Missing non-empty ENV for: ${ missingENVs.join(', ')` );
  }

  …
}
```

test/run/run-e2e-tests.js
```diff
@@ -0,0 +1,53 @@
...
+
+        await initTests();
+    }
+    catch ( err ) {
+        console.error( err );
+    }
+    finally {
+        process.exit();
+    }
+}
```
dmsnell: just because the entire function body is a `try` maybe it's worth splitting? probably fine as-is, but feels a little silly having a big function body inside a `try`

```js
async function run_unsafe() {
  …
}

async function run() {
  try {
    run_unsafe();
  } catch (err) {
    console.error( err );
  } finally {
    process.exit()
  }
}
```

.circleci/config.yml
```diff
@@ -396,18 +396,7 @@ jobs:
...
-                  DRIVERLOG=$(echo "$LOGDIR/chromedriver.log")
-
-                  pushd "release/mac/WordPress.com.app/Contents/MacOS" 
-                  eval './WordPress.com --disable-renderer-backgrounding --disable-http-cache --start-maximized --remote-debugging-port=9222 > "$APPLOG" 2>&1 &' && sleep 5
-                  popd
-                  
-                  ./node_modules/.bin/chromedriver --port=9515 --verbose > "$DRIVERLOG" 2>&1 &
-                  ./node_modules/.bin/mocha test/tests/e2e.js
-                  osascript -e 'quit app "WordPress.com"'
+                  make e2e
```
dmsnell: should we just call `npm run e2e` here instead of calling through `make`?

test/run/init_tests.js
```diff
@@ -0,0 +1,22 @@
...
+            if ( error ) {
+                reject( error );
+                return
+            } else {
+                resolve( stdout? stdout : stderr );
+            }
+        } );
+        tests.stdout.pipe( process.stdout );
+        tests.stderr.pipe( process.stderr );
+    } );
```
dmsnell: given my comment about the process runner, this one almost seems like it would be fine inline in the main test runner module. it's ugly, but if we remove some of the variables here it's not so bad, especially since we remove a level of `Promise`

```js
async function run() {
  …

  // run the tests
  const testPath = path.join( __dirname, '../../test/tests/e2e.js' );

  // by default should pipe stdout and stderr to this parent process
  // will throw an error on exception
  execSync( `npx mocha ${ testPath } --timeout 20000` );
}
```

test/run/checks.js
nsakaimbo: > ... was there something odd about the condition that prevented us from casting it to a boolean?

Nope - just my lack of experience writing idiomatic JavaScript. 😆Thanks for the suggestion - however casting to `Boolean` is now unnecessary as I went ahead and used the suggested credentials check refactor in [your comment](https://github.com/Automattic/wp-desktop/pull/778#discussion_r381594069) below, which works pretty well.

test/run/run-e2e-tests.js
nsakaimbo: Agree - this snippet of code isn't intuitive at first glance. However in this specific case, all the snippet is doing is some very minor re-formatting and a comment to clarify is sufficient (add [here](https://github.com/Automattic/wp-desktop/pull/778/files#diff-95f02e568cfd6b3a0b1830f8fd2306a1R65)):

```
// Replace `:` with `-` to format timestamp as YYYY-MM-DDTHH-MM-SS.mmmZ
```

I think a comment here is perfectly fine and creating an entire method to replicate all desired format components would add unnecessary complexity (and verbosity) to the file. If we choose to have a more customized date format in the future, we can create a separate method at that point.

test/run/init_app.js
nsakaimbo: This was in anticipation of Windows-specific syntax, however I've been able to determine that will be unnecessary and we can stick with `./WordPress.com`, which is nice.

test/run/init_app.js
nsakaimbo: > There's a part of me feeling like we could make a simple abstraction over running the process and then eliminate these two init… files

Agreed - having the various components initialized in separate files creates unnecessary indirection and overhead when navigating the script. I refactored to inline everything, and the proposed refactor for the `driver` and `app` initialization is a great idea. 👍 See refactor [here](https://github.com/Automattic/wp-desktop/pull/778/commits/c1862f5e56cb6e69217fd9f2110621afc1af3fcf#diff-95f02e568cfd6b3a0b1830f8fd2306a1R87).

test/run/run-e2e-tests.js
nsakaimbo: Agreed - see my response to your prior comment. Everything's been refactored inline. Also, I like the above as a much more concise way to check the required environment variables. See refactor [here](https://github.com/Automattic/wp-desktop/pull/778/files#diff-95f02e568cfd6b3a0b1830f8fd2306a1R59).

test/run/run-e2e-tests.js
nsakaimbo: In this specific case, the entirety of the logic directly encapsulated within the `try` block is evident at a glance, and further encapsulation in a wrapper method is unnecessary (and IMO adds an unnecessary level of indirection, and could be even be confusing). My preference here is to leave it as-is.

.circleci/config.yml
nsakaimbo: I think it might be nice to have the option of calling the tests either way. There's also some benefit to exposing/documenting critical workflow tooling in the makefile. In addition, all paths in the test script are absolute, which, combined with `make`, gives the developer flexibility in how they are invoked. (e.g. tweaking something from within the Calypso submodule and invoking with `make -f ./Makefile e2e`).

test/run/init_tests.js
nsakaimbo: Great suggestion. Refactored inline with `execSync` [here](https://github.com/Automattic/wp-desktop/pull/778/files#diff-95f02e568cfd6b3a0b1830f8fd2306a1R84). We have to add the additional `stdio: 'inherit` option so we can leverage streamed (and not buffered) output to the current shell, and it works great. 👍

test/run/run-e2e-tests.js
dmsnell: the comment is helpful enough since it shows what you wanted. my point was made though since I didn't expect `.mmmZ` 😄

test/run/run-e2e-tests.js
```diff
@@ -0,0 +1,94 @@
...
+    app.on( 'error', err => {
+        throw `failed to initialize command "${ command }": "${ err }"`;
+    } );
+    return app;
+}
+
+function initLogs( timestamp ) {
+    const dir = path.join( PROJECT_DIR, 'test', 'logs', `${ timestamp }` );
+
+    if ( !existsSync( dir ) ) {
```
dmsnell: note that this introduces a race condition that may not protect us. the directory could be created between the call to `exists` and the call to `mkdir`

probably simpler to just call `mkdirSync` and handle failure if the directory already exists. it may only fail in windows.

in other words, for the same level of safety we can skip `existsSync()`

```js
try {
	mkdirSync( dir, { recursive: true } );
} catch (e) {
	// pass
}
```

or do the more elegant/complete thing

```js
try {
	mkdirSync( dir, { recursive: true } );
} catch (e) {
	if (!existsSync(dir)) {
		console.error('Could not create log directory');
		process.exit(1);
	}
}
```

dmsnell: According to the docs it may never throw an error if `recursive: true` and we're not creating `/` on Windows

> Calling fs.mkdir() when path is a directory that exists results in an error only when recursive is false.

so in that case we may have the same safety as with `existsSync()` without any further try/catch. if it fails the build crashes as it must.

```js
…

mkdirSync(dir, {recursive: true });

…
```

test/run/run-e2e-tests.js
```diff
@@ -0,0 +1,94 @@
...
+        ], driverLog );
+
+        const tests = path.join( PROJECT_DIR, 'test', 'tests', 'e2e.js' );
+        execSync( `npx mocha ${ tests } --timeout 20000`, { stdio: 'inherit' } );
+    }
+    catch ( err ) {
+        console.error( err );
+    }
+    finally {
+        process.exit();
```
dmsnell: it's probably worth a comment here explaining that this kills the child processes which otherwise might continue to run. looking at the code it's odd seeing `finally { process.exit() }` as the last real line of execution. someone might come around and "improve" the code by removing it

nsakaimbo: > ... someone might come around and "improve" the code by removing it

😆Good point. Added comment in  87b55d5.

test/run/run-e2e-tests.js
nsakaimbo: Thanks for double-checking the usage here. Simplified this in f4021b9

PR #782, opened 2020-02-24
Revert workaround for buggy after_sign_hook

loremattei:
### Description:
This PR reverts a workaround for a bug in `electron-builder` where the `after_sign_hook` was called before the app was signed. 

### Motivation and Context:
https://github.com/Automattic/wp-desktop/pull/775 updated `electron-builder` to a version where the original issue has been fixed. 
Also, we found that the workaround doesn't work well for signing the `.zip` file.

### How Has This Been Tested:
1. Verify that the build on the CI to be green.
2. Tag the head of this branch with a new v5 beta tag and verify that the build is completed, the release on GitHub drafted and the .zip and .dmg binaries signed.

### DIFF
--- a/after_sign_hook.js
+++ b/after_sign_hook.js
@@ -3,14 +3,11 @@ const path = require('path');
 var electron_notarize = require('electron-notarize');
 
 module.exports = async function(params) {
-	// Only notarize the app on Mac OS only.
-	for (var key of params.platformToTargets.keys()) {
-		if (key.name !== 'mac') {
-			return;
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


nsakaimbo:
Looks good - thanks, @loremattei. Just a thought - do you think it might be helpful to test this change with a temporary `v1.0.0` tag (which we can subsequently delete afterwards)? Or are we fairly confident this will address the `mac.zip` notarization issue?

loremattei:
Sure @nsakaimbo! I was just writing the description :-)

PR #785, opened 2020-02-25
[e2e] Add Purchase flow

Stojdza:
### Description:

E2E tests - Add Business plan purchase to existing Sign up flow. Since we didn't succeed to set up sandbox payment cookie and the actual purchase couldn't be tested, I add the flow and instead of purchase, the test will empty the cart and check if user is logged in and signed up successfully. 

### Motivation and Context:

Expand the current set of e2e tests for wp-desktop app. 

### How Has This Been Tested:

Make sure that tests are passing in CircleCI

### Todos:

If we manage to set payment cookie for wp-desktop app, this flow should include actual purchase of, ie. Business plan.


test/tests/e2e.js
```diff
@@ -119,8 +121,7 @@ describe( 'Can Log Out', function() {
 describe( 'Can Sign up', function() {
 	this.timeout( 30000 );
 	const blogName = dataHelper.getNewBlogName();
-	const expectedBlogAddresses = dataHelper.getExpectedFreeAddresses( blogName );
-	const emailAddress = blogName + process.env.E2E_MAILOSAUR_INBOX;
```
nsakaimbo: Is the `E2E_MAILOSAUR_INBOX` environment variable no longer necessary? If so, we should remove from the README as well as the test running script.

nsakaimbo:
> If we manage to set payment cookie for wp-desktop app, this flow should incule actual purchase of, ie. Business plan

@Stojdza Thanks for your hard work in adding this to the test suite. Perhaps we can create a Github issue for this item so we can revisit down the road?

test/tests/e2e.js
Stojdza: Done in https://github.com/Automattic/wp-desktop/pull/785/commits/d15fdb36e6fd5a9adc5a8c3302af9db6eb7f30af

It's not necessary since we don't validate email after signing up. So we don't need a valid email address.

nsakaimbo: We'd need to remove the reference to the mailbox env var here as well:

https://github.com/Automattic/wp-desktop/blob/569952987fec4e64fdf2d3e95bbe84327a6f8a2d/test/run/run-e2e-tests.js#L58

Stojdza:
@nsakaimbo As already discussed in Slack - I've updated Sign up test so it will finish if the Checkout page if displayed properly. Since there is an AB test behind Checkout page and we don't handle AB tests in wp-desktop, we're seeing two different layouts of this page which make test unreliable to ontinue further. 

I'll add the following issues to the backlog:
- Handle AB tests, so we are always seeing default screens;
- Finish Checkout, once they remove AB test behind it;
- Set up Sandbox Payment Cookie so we can test actual purchase.

test/tests/e2e.js
Stojdza: Done in https://github.com/Automattic/wp-desktop/pull/785/commits/335f4c7db6090db1d5eca0286a73fb19908529b4

Stojdza:
I've removed `E2E_MAILOSAUR_INBOX` from CircleCI env vars

PR #789, opened 2020-02-27
File Loader: Add .mp4 Extension

Aurorum:
### Description:
This adds the `.mp4` file extension to the Webpack File Loader. 

### Motivation and Context:

For more details, see Automattic/wp-calypso#39707, but essentially, more videos are going to be uploaded to Calypso; this would allow them to be stored in the repo. 

### How Has This Been Tested:

In all honesty, I'm not really sure how testing works with the Desktop app, but it is just a minor change! 🙂

### DIFF
--- a/webpack.shared.js
+++ b/webpack.shared.js
@@ -48,7 +48,7 @@ module.exports = {
 				loader: 'ignore-loader',
 			},
 			{
-				test: /\.(?:gif|jpg|jpeg|png|svg)$/i,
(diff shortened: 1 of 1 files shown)


mmtr:
After some internal discussion (p4TIVU-9n1-p2), we finally decided that our marketing team will be making these videos and uploading them to our [YouTube channel](https://www.youtube.com/channel/UCvP5cfpfIgkb0Dt8o7Qt21A), so we won't need the loader to support .mp4 for now.

Issue #792, opened 2020-02-28
Unable to Use Mac App Artifact in "app.zip" Folder


nsakaimbo:
Note: this doesn't affect the .DMG installer, but is limited to the compressed .zip created by the packager and published to Github.

### Description

We recently discovered that the app embedded in the "app.zip" update artifact cannot be run on Catalina, with the message: "The application WordPress.com can't be opened". Turns out this is a known/oustanding issue related to code-signing with electron-builder:

github/electron-userland/electron-builder/issues/4299

It appears the root cause of this bug is that:

> If however you compress the .app file, then run the ./node_modules/app-builder-bin/mac/app-builder blockmap -i dist/[YourApp]-${version}-mac.zip command, this will modify the .zip file, and this file will then extract properly on Catalina, but the inner .app file will not run.
> 
> You can see that running the blockmap command many times will yield different signatures each time.

(from [this](https://github.com/electron-userland/electron-builder/issues/4299#issuecomment-553484330) comment)

Inspecting the application with `codesign --test-requirement="=notarized" --verify --verbose <appname.app>` confirms the extracted .app is not codesigned at all.

### Problem

At present, this doesn't appear to be an immediate concern with respect to the auto-update mechanism (it's unclear the usage of both the .zip and the .dmg during the auto-update process, and we need to understand this in more detail).

We will soon need to address Issue #790 (i.e. address a security vulnerability by dropping the `Disable Library Validation` entitlement and instead ensure that all embedded third-party executables are code-signed within the notarized .app bundle). If the auto-updater does in fact uses assets from the .zip file, the extracted assets _not_ being codesigned will likely be a blocker at that point (i.e. once the `Disable Library Validation` entitlement is dropped, Catalina will not allow the application to run unless all embedded executables are codesigned).

nsakaimbo:
Turns out this is a non-issue. This is a [bug in Mac OS 10.15](https://github.com/electron-userland/electron-builder/issues/4299#issuecomment-544683923) resulting from extraction invoked by double-clicking a zip archive.

The bundle when auto-extracted after downloading via Safari (not Chrome) is confirmed to be signed and notarized correctly:

`codesign --test-requirement="=notarized" --verify --verbose WordPress.com.app`

To avoid this scenario, use `ditto` if manually extracting the .app from the zip archive:

`ditto -x -k <zip archive> <destination folder>`

PR #797, opened 2020-03-09
[Notarization] Streamline Notarization Logic

nsakaimbo:
### Description:

Some minor cleanup of notarization logic.

Per the current [Electron docs](https://www.electron.build/configuration/dmg), DMG signing is disabled by default as it will "lead to unwanted errors in combination with notarization requirements." Only the app bundle is zipped and submitted to Apple for notarization. Also worth pointing out that `electron-notarize` does not support DMG notarization anyway, so the extraneous logic should be removed.

This PR removes the unnecessary (and unused) call to submit the DMG for notarization, and adds some notes to the README regarding correct extraction of the notarized app from the published .zip artifact.

### Motivation and Context:

As a result of investigating the issue reported in #792, a deeper dive into existing notarization logic was conducted and some efficiencies/refactors were identified. Turns out #792 is a non-issue (this is a bug in Mac OS 10.15 resulting from default extraction invoked by double-clicking a zip archive. Instead, use `ditto` as specified in the README).

### How Has This Been Tested:

Tested by submitting tagged test release to Apple.

[Before](https://circleci.com/gh/Automattic/wp-desktop/54504):

```
 • signing         file=release/mac/WordPress.com.app identityName=Developer ID Application: Automattic, Inc. (XXXXX) identityHash=XXXXX provisioningProfile=none
afterSign hook triggered

// ...

Notarizing com.automattic.wordpress found at /Users/distiller/wp-desktop/release/mac/WordPress.com.app
Done notarizing com.automattic.wordpress
  • building        target=macOS zip arch=x64 file=release/WordPress.com-5.0.0-mac.zip
  • building        target=DMG arch=x64 file=release/WordPress.com-5.0.0.dmg
  • building embedded block map  file=release/WordPress.com-5.0.0-mac.zip
  • publishing      publisher=Github (owner: Automattic, project: wp-desktop, version: 5.0.0)
  • uploading       file=WordPress.com-5.0.0-mac.zip provider=GitHub

// Note: DMG notarization is not triggered (but is unnecessary anyway)
```

 After:

```
  • signing         file=release/mac/WordPress.com.app identityName=Developer ID Application: Automattic, Inc. (XXXXX) identityHash=XXXXX provisioningProfile=none
  • notarizing WordPress.com.app...
  • done notarizing WordPress.com.app, took 4 minutes, 1 seconds
  • building        target=DMG arch=x64 file=release/WordPress.com-1.0.0.dmg
  • building block map  blockMapFile=release/WordPress.com-1.0.0.dmg.blockmap
```

### DIFF
--- a/after_sign_hook.js
+++ b/after_sign_hook.js
@@ -1,46 +1,45 @@
-const fs = require('fs');
-const path = require('path');
-var electron_notarize = require('electron-notarize');
+#!/usr/bin/env node
 
-module.exports = async function(params) {
-  // Only notarize the app on Mac OS only.
-  if (params.electronPlatformName !== 'darwin') {
-    return;
-  }
-  console.log('afterSign hook triggered', params);
+const path = require( 'path' );
+const { notarize } = require( 'electron-notarize' );
 
-  if (!process.env.CIRCLE_TAG || process.env.CIRCLE_TAG.length === 0) {
-    console.log('Not on a tag. Skipping notarization');
-    return;
-  }
+const APP_ID = 'com.automattic.wordpress';
+const NOTARIZATION_ASC_PROVIDER = 'AutomatticInc';
+const NOTARIZATION_ID = process.env.NOTARIZATION_ID;
+const NOTARIZATION_PWD = process.env.NOTARIZATION_PWD;
 
-  // Same appId in electron-builder.
-  let appId = 'com.automattic.wordpress';
+const shouldNotarize = process.platform === 'darwin' && !!process.env.CIRCLE_TAG;
 
-  let appPath = params.appOutDir
-    ? path.join(
-        params.appOutDir,
-        `${params.packager.appInfo.productFilename}.app`
-      )
-    : params.artifactPaths[0].replace(new RegExp('.blockmap'), '');
+function elapsed( start ) {
+	const now = new Date();
 
-  if (!fs.existsSync(appPath)) {
-    throw new Error(`Cannot find application at: ${appPath}`);
-  }
+	const ms = Math.abs( now.getTime() - start.getTime() );
+	const diff = new Date( ms );
 
-  console.log(`Notarizing ${appId} found at ${appPath}`);
+	return `${ diff.getMinutes() } minutes, ${ diff.getSeconds() } seconds`;
+}
 
-  try {
-    await electron_notarize.notarize({
-      appBundleId: appId,
-      appPath: appPath,
-      appleId: process.env.NOTARIZATION_ID,
-      appleIdPassword: process.env.NOTARIZATION_PWD,
-      ascProvider: 'AutomatticInc',
-    });
-  } catch (error) {
-    console.error(error);
-  }
+module.exports = async function( context ) {
+	if ( ! shouldNotarize ) {
+		return;
+	}
 
-  console.log(`Done notarizing ${appId}`);
-};
+	try {
+		const app = path.join( context.appOutDir, `${ context.packager.appInfo.productFilename }.app` );
+		const appName = path.basename( app );
+
+		const start = new Date();
+		console.log( `  • notarizing ${ appName }...` );
+		await notarize( {
+			appBundleId: APP_ID,
+			appPath: app,
+			appleId: NOTARIZATION_ID,
+			appleIdPassword: NOTARIZATION_PWD,
+			ascProvider: NOTARIZATION_ASC_PROVIDER,
+		} )
+		console.log( `  • done notarizing ${ appName }, took ${ elapsed( start ) }` );
+	} catch ( error ) {
+		console.log( error.message );
+		throw ( error );
+	}
+}
--- a/after_sign_hook.js
+++ b/after_sign_hook.js
@@ -24,10 +24,10 @@ module.exports = async function( context ) {
 		return;
 	}
 
-	try {
-		const app = path.join( context.appOutDir, `${ context.packager.appInfo.productFilename }.app` );
-		const appName = path.basename( app );
+	const app = path.join( context.appOutDir, `${ context.packager.appInfo.productFilename }.app` );
+	const appName = path.basename( app );
 
+	try {
 		const start = new Date();
 		console.log( `  • notarizing ${ appName }...` );
 		await notarize( {
--- a/after_sign_hook.js
+++ b/after_sign_hook.js
@@ -39,7 +39,6 @@ module.exports = async function( context ) {
 		} )
 		console.log( `  • done notarizing ${ appName }, took ${ elapsed( start ) }` );
 	} catch ( error ) {
-		console.log( error.message );
 		throw ( error );
 	}
 }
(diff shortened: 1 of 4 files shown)


after_sign_hook.js
```diff
@@ -1,46 +1,45 @@
...
+		const start = new Date();
+		console.log( `  • notarizing ${ appName }...` );
+		await notarize( {
+			appBundleId: APP_ID,
+			appPath: app,
+			appleId: NOTARIZATION_ID,
+			appleIdPassword: NOTARIZATION_PWD,
+			ascProvider: NOTARIZATION_ASC_PROVIDER,
+		} )
+		console.log( `  • done notarizing ${ appName }, took ${ elapsed( start ) }` );
```
dmsnell: this `elapsed()` may be more complicated than we need it to be.

do we expect it to take many minutes? would "took 2590s" be too unhelpful?
could also consider using a common idiom, `humanTime()` or `humanDiff()`
while keeping the rest easy so that we're not mingling elapsed time and display

```js
const start = Date.now(); // returns seconds
…
const finish = Date.now();
console.log( ` • done notarizing ${ appName }, took ${ ( finish - start ) / 1000 } seconds` );
// or
console.log( ` • done notarizing ${ appName }, took ${ humanTime( finish - start ) } seconds` );

/**
 * We don't expect to take hours
 */
function humanTime( ms ) {
  // convert to time string
}
```

Probably I'm guessing that the human time isn't necessary, that seconds are probably enough - what are they used for? For instance, my shell already tells me how long the last command took if it takes more than 10 seconds; won't CircleCI report runtime as well?

In summary:
 - it seems like we're going out of our way to provide this
 - `elapsed` seems to cross both abstractions of time and display (what does it mean if we call `diff.getYear()` and it returns `69`?) when it could instead focus on just display while keeping the time element simpler, e.g. by using `Date.now()` to return seconds

after_sign_hook.js
```diff
@@ -1,46 +1,45 @@
...
+}
+
+module.exports = async function( context ) {
+	if ( ! shouldNotarize ) {
+		return;
+	}
+
+	try {
+		const app = path.join( context.appOutDir, `${ context.packager.appInfo.productFilename }.app` );
+		const appName = path.basename( app );
```
dmsnell: we should probably move these outside of the `try` since we don't expect them to fail, yeah?

after_sign_hook.js
```diff
@@ -1,46 +1,45 @@
...
+			appPath: app,
+			appleId: NOTARIZATION_ID,
+			appleIdPassword: NOTARIZATION_PWD,
+			ascProvider: NOTARIZATION_ASC_PROVIDER,
+		} )
+		console.log( `  • done notarizing ${ appName }, took ${ elapsed( start ) }` );
+	} catch ( error ) {
+		console.log( error.message );
+		throw ( error );
+	}
```
dmsnell: since all we're doing is throwing the error and printing it then is there any reason to catch it? what happens if we don't? won't it still print the error and throw it, doing almost exactly what it does here, or does it not print out the `error.message` because some higher-up code is catching the error and hiding it?

after_sign_hook.js
nsakaimbo: > do we expect it to take many minutes? would "took 2590s" be too unhelpful?

and

> won't CircleCI report runtime as well?

Notarization takes anywhere in the ballpark of ~5 minutes to an hour ( per Apple docs), and I have seen this metric vary significantly within that range. Circle CI provides the total runtime of a given job, but doesn't provide info about the time when console output within a job is updated. Outputting this info as minutes-and-seconds is immensely helpful and makes it more immediate/intuitive to grok, so I'm inclined to keep the format as-is.

> it seems like we're going out of our way to provide this

The turnaround of Apple's servers is useful to know on an ongoing basis, and tracking this stat will be informative in evaluating any significant deviation from the norm.

after_sign_hook.js
nsakaimbo: It looks a little nicer when output to console, but agree it's extraneous - removed in dbd8b42.

after_sign_hook.js
dmsnell: > Outputting this info as minutes-and-seconds is immensely helpful and makes it more immediate/intuitive to grok, so I'm inclined to keep the format as-is.

👍 

> The turnaround of Apple's servers is useful to know on an ongoing basis, and tracking this stat will be informative in evaluating any significant deviation from the norm.

Should this be reported in computer-readable format somewhere? Fire off a stat automattically?

after_sign_hook.js
dmsnell: I meant the whole `try/catch` block. Right now it's this…

```js
try {
  something()
} catch (anything) {
  throw (anything);
}
```

which I _think_ is equivalent to this…

```js
something();
```

after_sign_hook.js
nsakaimbo: Yep - makes sense. In the unlikely event these constants are undefined the logic an error will be thrown any way. Updated in bb56851

after_sign_hook.js
nsakaimbo: tbh was simply unclear what the idiomatic way to "rethrow" in JavaScript (or whether it should be done at all). Seems there is a subtle [difference in stack trace](https://stackoverflow.com/a/49323141) between a throw and rethrow, but doesn't seem suited to this use case. I think doing away with the try/catch here is fine - 17253be6.

after_sign_hook.js
nsakaimbo: > Should this be reported in computer-readable format somewhere? Fire off a stat automattically?

Yeah that's a good thought, and might make this more useful down the road. I'd certainly like to keep that in mind for a potential future enhancement. For the time being we can manually look up the build logs for any given tag (since we only notarize on tag) and make an assessment that way (if and when we need do).

loremattei:
Thanks for diving into it @nsakaimbo!

`dmg` notarisation has been discussed quite a lot in the first days and I think some changes were made around `10.14.5` to fix some issues.  

I tested the `dmg` built with this branch (downloaded from CircleCI) and I get this message when I try to run the app:
Screenshot 2020-03-12 11 42 10

While it's still possible to run the app right clicking on the icon, I wonder if this is adding confusion for the users. 

If I try with the released `5.0.0` DMG from GitHub, I get the standard warning with the `open` option: 
Screenshot 2020-03-12 11 42 46

I'm not totally sure what can cause this difference, but I think the current message is easier to handle for the users.

nsakaimbo:
@loremattei I believe you were testing artifacts not built on a tag, which means the app bundle wasn't notarized. I just built test artifacts with a v1.0.0 test tag and the DMG and app appear to behave as expected, but let me know if you still observe the same behavior: 

[link](https://circleci.com/gh/Automattic/wp-desktop/55778#artifacts/containers/0)

loremattei:
👌I tested the DMG in the link and it looks good!
:shipit:

PR #802, opened 2020-03-10
[e2e] Disable one check

Stojdza:
Disable one check for composite checkout. I'm still not sure why previous failures have happened ([example](https://circleci.com/gh/Automattic/wp-desktop/55550#tests/containers/0)), video of failed test looks as expected. 

This is a temporary solution while we don't find the root cause. @nsakaimbo if Sign-up test continues to fail, feel free to disable it, and I will investigate as soon as I grab some time :)


nsakaimbo:
Update: tests passed in two consecutive runs of this branch -- merging.

PR #807, opened 2020-03-19
[Security] Update Electron Builder

### LINKED ISSUE #791: [Vulnerability] Updater Verification Bypass (Windows)
### Description

A vulnerability has been reported that could lead to a Signature Validation Bypass of the updates mechanism in electron-updater, exposing affected clients to an increased risk of Remote Code Execution.

* The signature verification check performed by electron-updater on Wordpress Desktop for Windows is simply based on a string comparison between the binary’s publisherName and the hardcoded value defined in the latest.yml file.

* An attacker can bypass the entire signature verification by triggering a parse error in the script responsible for parsing the filename. This can be achieved by using a filename containing a single quote.

[reported via Hacker One]

### Remediation

When available, merge v22.3.5 of electron-builder (currently in pre-release status) which includes a fix for this issue.

nsakaimbo:
### Description

This PR updates `electron-builder` to v22.4.0. This version bump includes a patch made in v22.3.5 that addresses a security vulnerability on Windows.

Ref: electron-userland/electron-builder@dead150

_Minor Update to Makefile_

Turns out the `BUILD_PLATFORM` environment variable was not being set in the Windows CI job (so the Makefile was resolving as `electron-builder build -<BLANK>` on Windows) and the latest `electron-builder` will complain about the missing platform identifier. Since the platform identifier is only necessary if building in a cross-platform manner, we can do away with it entirely. (By default, `electron-builder build` will build for the platform it is invoked on.)

Fixes #791

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -281,7 +281,7 @@ jobs:
           command: |
                   source $HOME/.nvm/nvm.sh
                   nvm use
-                  make package BUILD_PLATFORM=l
+                  make package
       - run:
           name: Release cleanup
           command: |
@@ -382,7 +382,7 @@ jobs:
                   set +e
                   source $HOME/.nvm/nvm.sh
                   nvm use
-                  make package BUILD_PLATFORM=m
+                  make package
       # Temporarily disable the test
       # TODO: Fixme – seems to be an issue with `portscanner`
       # - run:
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,6 @@ CHECKMARK = OK
 CONFIG_ENV = 
 CALYPSO_ENV = desktop
 NODE_ENV = production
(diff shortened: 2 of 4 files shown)


loremattei:
cc/ @rachelmcr as updating `electron-builder` caused some issues with the auto updater in the past. 
We are now on much more recent versions, so I don't think this is going to be the case, but it's probably worth to pay a bit of extra attention to any update madness during our testing :-)

nsakaimbo:
> updating electron-builder caused some issues with the auto updater in the past

Good to know - thanks for the heads up! We should probably focus a little more on verifying auto-update when we move to beta with this code change.

rachelmcr:
Thanks for the heads up! cc @designsimply (who will be wrangling the calls for testing)

PR #815, opened 2020-03-23
feat(logging): replace debug with custom logger

nsakaimbo:
# [2] Replace Debug With Custom Logger

## Description

This PR is a follow-on to #814 and replaces the `debug` module with the custom logger implementation. Existing logging items are converted from the usage of `debug` to `lib/logger`.

## Highlights

Conversion of existing log items is fairly straightforward. For the preload script, however, log items are transmitted via IPC from the renderer to the main process. This is to avoid use of the `remote` module, which is regarded as highly insecure (and in fact will be excluded from future versions of Electron starting with v9).

## To Test

Build and run the application (or download [built artifacts for this PR](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/9418/workflows/a024e7b4-6b65-4d67-a304-f64b5c8d1b11/jobs/56585)). Verify application logs are persisted to the following locations on each platform:

_Note_: Mac artifact is not notarized (notarized on tag only), so will require opening via System Preferences -> Security.

  - Mac: `$HOME/Library/Application Support/WordPress.com/logs`
  - Linux: `/home/<username>/.config/WordPress.com/logs`
  - Windows 10: `C:\Users\<username>\AppData\Roaming\WordPress.com\logs`

### DIFF
--- a/desktop/app-handlers/logging/ipc-handler/index.js
+++ b/desktop/app-handlers/logging/ipc-handler/index.js
@@ -1,0 +1,59 @@
+'use strict';
+
+/**
+ * External Dependencies
+ */
+const { ipcMain: ipc } = require( 'electron' );
+
+module.exports = {
+    /**
+     * JSON object that stores references to loggers initialized
+     * by calls from the renderer process over IPC (keyed by `namespace`).
+     */
+	_loggers: {},
+
+    /**
+     * Gets the logging object for the provided `namespace` (and creates a logger
+     * if one doesn't exist). Logger references are tracked in the private
+     * `_loggers` object and keyed by `namespace`.
+     * @param {String} namespace Namespace of the logger to be used or initialized.
+     * @param {Any} options Logger configuration.
+	 * @returns {Object} Logger instance.
+     */
+	_getLogger: function( namespace, options ) {
+		let logger = this._loggers[namespace];
+		if ( !logger ) {
+			logger = require( 'lib/logger' )( namespace, options );
+			this._loggers[namespace] = logger;
+		}
+		return logger;
+	},
+
+    /**
+     * Registers an ipc listener on the `log` channel to relay logging
+     * events from the renderer process.
+     */
+	listen: function() {
+		ipc.on( 'log', ( _, level, namespace, options, message, meta ) => {
+			const logger = this._getLogger( namespace, options );
+
+			switch ( level ) {
+				case 'error':
+					logger.error( message, meta );
+					break;
+				case 'warn':
+					logger.warn( message, meta );
+					break;
+				case 'info':
+					logger.info( message, meta );
+					break;
+				case 'debug':
+					logger.debug( message, meta );
+					break;
+				case 'silly':
+					logger.silly( message, meta );
+					break;
+			}
+		} )
+	}
+}
--- a/desktop/app-handlers/logging/ipc-handler/index.js
+++ b/desktop/app-handlers/logging/ipc-handler/index.js
@@ -12,14 +12,14 @@ module.exports = {
      */
 	loggers: {},
 
-    /**
-     * Gets the logging object for the provided `namespace` (and creates a logger
-     * if one doesn't exist). Logger references are tracked in the private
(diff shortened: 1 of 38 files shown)


desktop/app-handlers/logging/ipc-handler/index.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
```
nsakaimbo: **Callout**: We use a handler in the main process to listen for log items transmitted from the renderer process.

public_desktop/preload.js
```diff
@@ -3,22 +3,41 @@ const { SpellCheckHandler, ContextMenuListener, ContextMenuBuilder } = require(
...
 require( 'rxjs/add/observable/from' ); // necessary for providing Observable.from() to Observable (see usage below)
 
-const debugModule = require( 'debug' );
+// FIXME: The `remote` module is being deprecated and will be excluded
+// from Electron v9.x.
 
+// It is possible to use the `remote` module to directly require the
+// main process logging module. However, this is a security concern and is
+// discouraged. Instead, we post log messages to the main process via ipc.
+const send = ( level, namespace, options, message, meta ) => {
```
nsakaimbo: **Callout**: send log messages from the renderer via IPC.

desktop/app-handlers/logging/ipc-handler/index.js
```diff
@@ -0,0 +1,59 @@
...
+ * External Dependencies
+ */
+const { ipcMain: ipc } = require( 'electron' );
+
+module.exports = {
+    /**
+     * JSON object that stores references to loggers initialized
+     * by calls from the renderer process over IPC (keyed by `namespace`).
+     */
+	_loggers: {},
```
dmsnell: quick note: we don't use the underscore to indicate private methods. [Coding guidelines](https://wpcalypso.wordpress.com/devdocs/docs/coding-guidelines/javascript.md)

It's not entirely or clearly spelled out, but in the React component section it suggests that
we avoid method names prefixed with `_`. In general we don't use too many private methods
at all.

nsakaimbo: Thanks for the tip, definitely something I was unaware of. Updated in 6649f35.

desktop/app-handlers/updater/manual-updater/index.js
```diff
@@ -39,8 +39,8 @@ class ManualUpdater extends Updater {
 	async ping() {
 		try {
 			const url = this.apiUrl;
-			debug( 'Checking for update. Fetching:', url );
-			debug( 'Checking for beta release:', this.beta )
+			log.info( 'Checking for update. Fetching: %s', url );
```
belcherj: Why use this to build the string? You haven't used this elsewhere. Would a template literal make more sense?

desktop/app-handlers/logging/ipc-handler/index.js
```diff
@@ -0,0 +1,59 @@
...
+     */
+	loggers: {},
+
+    /**
+     * Gets the logging object for the provided `namespace` (and creates a logger
+     * if one doesn't exist). Logger references are tracked in the private
+     * `loggers` object and keyed by `namespace`.
+     * @param {String} namespace Namespace of the logger to be used or initialized.
+     * @param {Any} options Logger configuration.
+	 * @returns {Object} Logger instance.
```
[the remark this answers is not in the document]
nsakaimbo: hrm - not sure what happened. Fixed in 46d8da63a27f14f19eb011cd989503309efc31d3.

desktop/app-handlers/updater/manual-updater/index.js
nsakaimbo: @belcherj This is a great callout. Using `%s` and other such tokenizers is something we absolutely shouldn't have to do. I addressed this by simplifying the log formatter even further so the user doesn't have to care about what types are being passed in (via `util.format`, which essentially lets us expect the same behavior as `console.log`.) It works well! 🎉 

Fixed in d0553354fe922615cac17ee0739c0eeb35846bba

PR #816, opened 2020-03-23
feat(logging): archive and export logs [3]

nsakaimbo:
# [3] Export Application Logs to User's Desktop

<p align="center">
wp-desktop-export-logs-to-desktop
</p>

## Description

This PR is a follow-on to #814 and #815, and adds a `Get activity logs` item to the application's `Help` menu. When selected, this menu item exports the user's logs to their desktop and prompts them to share their logs when requesting help. This feature makes use of [`archiverjs`](https://github.com/archiverjs/node-archiver) for cross-platform compression.

## To Test

1. Build and run the application, or download one of the [built artifacts](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/9444/workflows/c9ca5cb0-5649-43de-9ab4-9f09bb06dcc2/jobs/56621/artifacts) from this PR
2. Export your activity logs via the `Help` menu, and confirm the archive has been saved to your desktop.

_Note_: Mac artifact is not notarized (notarized on tag only), so will require opening via System Preferences -> Security.

(See detailed installation instructions in [this](https://github.com/Automattic/wp-desktop/pull/816#issuecomment-606065206) comment.)

### DIFF
--- a/desktop/window-handlers/get-logs/index.js
+++ b/desktop/window-handlers/get-logs/index.js
@@ -43,8 +43,8 @@ module.exports = async function( window ) {
 		dialog.showMessageBox( window, {
 			type: 'info',
 			buttons: [ 'OK' ],
-			title: 'Logs saved to your desktop',
-			message: 'Logs saved to your desktop' +
+			title: 'Application logs saved to your desktop',
+			message: 'Application logs saved to your desktop' +
 					'\n\n' +
 					`${path.basename( file )}`,
 			detail: 'For help with an issue, please contact help@wordpress.com and share your logs.'
--- a/desktop/window-handlers/get-logs/index.js
+++ b/desktop/window-handlers/get-logs/index.js
@@ -19,9 +19,7 @@ const log = require( 'lib/logger' )( 'desktop:get-logs' );
  */
 const logPath = state.getLogPath();
 
-function pad( n ) {
-	return n < 10 ? '0' + n : n;
-}
+const pad = n => `${n}`.padStart( 2, '0' );
 
 const localDateTime = () => {
 	const now = new Date();
--- a/desktop/window-handlers/get-logs/index.js
+++ b/desktop/window-handlers/get-logs/index.js
@@ -1,0 +1,76 @@
+'use strict';
+
+/**
+ * External dependencies
+ */
+const path = require( 'path' );
+const { app, dialog } = require( 'electron' );
+
+/**
+ * Internal dependencies
+ */
+const state = require( 'lib/state' );
+const system = require( 'lib/system' );
+const { zipContents } = require( 'lib/archiver' );
+const log = require( 'lib/logger' )( 'desktop:get-logs' );
+
+/**
+ * Module variables
+ */
+const logPath = state.getLogPath();
+
+function pad( n ) {
+	return n < 10 ? '0' + n : n;
+}
+
+const localDateTime = () => {
+	const now = new Date();
+	return now.getFullYear() +
+		'-' +
+		pad( now.getMonth() + 1 ) +
+		'-' +
+		pad( now.getDate() ) +
+		'T' +
+		pad( now.getHours() ) +
+		'.' +
+		pad( now.getMinutes() ) +
+		'.' +
+		pad( now.getSeconds() ) +
+		'.' +
+		pad( now.getMilliseconds() );
+}
+
+module.exports = async function( window ) {
+	const onZipped = async () => {
+		await dialog.showMessageBox( window, {
+			type: 'info',
+			buttons: [ 'OK' ],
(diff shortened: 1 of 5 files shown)


desktop/window-handlers/get-logs/index.js
```diff
@@ -0,0 +1,79 @@
...
+const log = require( 'lib/logger' )( 'desktop:get-logs' );
+
+/**
+ * Module variables
+ */
+const logPath = state.getLogPath();
+
+function pad( n ) {
+	return n < 10 ? '0' + n : n;
+}
```
dmsnell: feel free to use the built-in method to clarify the intent

```js
const pad = n => `${n}`.padStart(2, '0');
```

desktop/window-handlers/get-logs/index.js
```diff
@@ -0,0 +1,79 @@
...
+			'Please contact help@wordpress.com and mention the error details below:' +
+			'\n\n' +
+			error.stack +
+			'\n\n' +
+			'System info: ' + JSON.stringify( system.getVersionData() )
+		)
+	}
+
+	try {
+		const timestamp = localDateTime();
```
dmsnell: This also works, and provides the additional certainty that the dates are referenced against UTC and won't be confused by any local timezone setting or override in the browser.

```js
const timestamp = new Date().toISOString();
```

desktop/window-handlers/get-logs/index.js
```diff
@@ -0,0 +1,79 @@
...
+		await dialog.showMessageBox( window, {
+			type: 'info',
+			buttons: [ 'OK' ],
+			title: 'Logs saved to your desktop',
+			message: 'Logs saved to your desktop' +
+				'\n\n' +
+				`${ path.basename( file ) }`,
+			detail: 'For help with an issue, please contact help@wordpress.com and share your logs.'
+		} )
+	}
```
dmsnell: fun fact: the curlies here aren't necessary. they are fine, but they are superfluous

```js
const onZipped = async ( file ) => await dialog.showMessageBox( window, {
  …
} );
```

dmsnell: also, if `dialog.showMessageBox` returns a `Promise` and this function isn't doing anything with it, it might be superfluous to also wrap this in `async` - same with `onError` below.

won't the runtime be identical to if we declared these as non-async?

```js
const onZipped = file => dialog.showMessageBox( … );
```

desktop/window-handlers/get-logs/index.js
nsakaimbo: TIL 🤦‍♂

Updated in 4b23342

desktop/window-handlers/get-logs/index.js
nsakaimbo: The intent was to have a localized timestamp (non-UTC), but with similar formatting. I tried scouring for solutions on this, and short of importing something like `momjentjs` (which would be overkill here) the only way I could find was to generate each component in this manner. Would be great to do this in a more concise way, though.

desktop/window-handlers/get-logs/index.js
nsakaimbo: Yep - the async/await here is unnecessary, updated in bee3aad. Also fixed the callback as it was misspecified (should return a function).

belcherj: You don't need the curlies on L56 and L68, Arrow functions allow for the following syntax:

`const onError = ( error ) => dialog.showMessageBox( window, {...});`

Although that means that calling onError returns the result of dialog.showMessageBox rather than undefined.

nsakaimbo:
@belcherj Thanks for taking a peek! I double-checked and the menu item is there and works. Steps:

1. Download the `mac.zip` artifacts from [this link]( https://app.circleci.com/pipelines/github/Automattic/wp-desktop/9824/workflows/2b3e9368-08e3-484c-a8ba-6e981dcbbd6a/jobs/56988/artifacts)
2. In your terminal, go to your Downloads folder
3. Extract with `ditto -x -k WordPress.com-5.0.0-mac.zip .`
4. Go to the Downloads folder in Finder, and double-click the extracted app.
5. Because this app is not notarized (not on a tag), it will fail to open when clicked. You will have to go to `System Preferences -> Security & Privacy` and under the "General" tab select `Open Anyway`
6. Verify that the "Get application logs" menu item is present and works.

<p align="center">
Screen Shot 2020-03-30 at 11 17 54 AM
</p>

desktop/window-handlers/get-logs/index.js
nsakaimbo: Remove unnecessary curlies in 5ae2f9bb927a1267b487945ecd96b0dcc639a135

PR #821, opened 2020-03-30
fix(auto-updater): ensure existing version is quit and replaced

nsakaimbo:
### Description

This PR fixes a bug in the Windows auto-update logic where - when the user clicked `Install and Restart` - the current version was not shut down correctly, and the auto-updater did not completely/correctly replace the prior version. A reference to the prior version would be left in `Add/Remove Programs` (requiring a fix in the user's Registry) and the application's shortcut would potentially be unresponsive when clicked.

This issue is relatively easy to replicate and has been reported by at least one user (zen-2796859).

Debugging this issue seems indicates a couple of root causes at play:

* Existing logic that backgrounds the application when the main window is closed (instead of respecting the auto-updater's directive to quit the app), and
* Faulty logic in the current version of `electron-updater`, ([ref](https://github.com/electron-userland/electron-builder/issues/4143))

Updating the backgrounding logic and electron-builder version to latest seems to correct this issue.

### To Test

1. Uninstall any existing WordPress.com installation on Windows. Verify that no entries for `WordPress.com` are in `Add/Remove Programs` (may need to use the registry fix in https://support.microsoft.com/en-us/help/243723/removing-invalid-entries-in-the-add-remove-programs-tool if so).
1. Download and install the Windows artifact built from this branch
1. This version is set to `v5.0.0` (same as `develop`), and after launch should detect and prompt for (v5.0.1-beta) which is currently available (this may take a couple minutes).
1. Click `Install & Restart` when prompted.
1. Verify auto-update happens successfully, and that there is only one entry for `WordPress.com` in `Add/Remove Programs`).

### DIFF
--- a/desktop/lib/app-quit/index.js
+++ b/desktop/lib/app-quit/index.js
@@ -10,12 +10,12 @@ function AppQuit() {
 }
 
 AppQuit.prototype.shouldQuitToBackground = function() {
-	if ( this.canQuit === false ) {
-		return true;
+	if ( this.canQuit ) {
+		this.canQuit = false;
+		return false;
 	}
 
-	this.canQuit = false;
-	return false;
+	return true;
 };
 
 AppQuit.prototype.allowQuit = function() {
--- a/desktop/lib/platform/windows/index.js
+++ b/desktop/lib/platform/windows/index.js
@@ -3,9 +3,7 @@
 /**
  * External Dependencies
  */
-const electron = require( 'electron' );
-const Tray = electron.Tray;
-const Menu = electron.Menu;
+const { Tray, Menu, app } = require( 'electron' );
 const debug = require( 'debug' )( 'platform:windows' );
 
 /**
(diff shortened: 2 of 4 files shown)


desktop/lib/app-quit/index.js
```diff
@@ -10,12 +10,12 @@ function AppQuit() {
...
-		return true;
+	if ( this.canQuit ) {
+		this.canQuit = false;
+		return false;
 	}
 
-	this.canQuit = false;
-	return false;
+	return true;
 };
```
loremattei, line 19: Sorry, I miss what we are changing here 🤔

nsakaimbo: Thanks for the review/testing, @loremattei!

> I miss what we are changing here 

You're right - in the case of `shouldQuitToBackground`, I think the code change winds up with the same result. However, I found it challenging to reason about the logic as-written.

In addition, on Windows, the application was _not_ closing the main window, leading to a duplicate installation of the app. There were two methods of concern around how the application handles backgrounding and the `close` event (which is emitted by `app.quitAndInstall`).

```
AppQuit.prototype.shouldQuitToBackground = function() {
	if ( this.canQuit === false ) {
		return true;
	}

	this.canQuit = false;
	return false;
};
```

```
WindowsPlatform.prototype.onClosed = function( ev ) {
	if ( appQuit.shouldQuitToBackground() ) {
		log.info( 'Window close puts app into background & creates tray' );

		ev.preventDefault();

		this.window.hide();
		this.showBackgroundBubble();
	}
};
```

I found the above a little difficult to reason about, and refactored things slightly to make sure that each code path is explicitly handled as well as make things a little easier to grok.

loremattei: 👍

PR #822, opened 2020-04-01
improvement(tooling): add Windows dev tooling

nsakaimbo:
### Description

To-date, debugging issues locally on Windows has been virtually non-existent (due to the fact that building the application source is not supported on Windows). When troubleshooting Windows-specific issues locally, a developer has to rely on artifacts built via Circle for each code change when attempting to debug. This PR adds some tooling to make building and debugging the application locally on Windows a little more convenient:

* Docker environment for building the application source
* Convenience Docker invocation added to the Makefile
* Documentation of recommended Windows environment

These changes facilitate being able to build both the packaged app and the developer environment locally. 🎉

The Docker environment is configured with NVM and all essential build tools, meaning that the image does not have to be re-built each time and can serve as a literal virtual Linux environment persistent on the user's Windows installation. 

### To Test

1. Set up Windows environment (MSYS2/Git Bash), Docker, windows-build-tools
2. Build the Docker image with `make docker-build`. Invoke with `make docker-run`. Clean with `make docker-clean`.
3. As with CI, the application source must be built in Linux, but packaged natively in Windows. 

_Instructions_

* To build the packaged (i.e. "final") app artifacts, run `make build-source` within the Docker environment, and `make package` on Windows. 
* To run the application in developer mode, execute `make dev-server` in Docker and `make dev` on Windows.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -334,7 +334,7 @@ jobs:
                 Import-PfxCertificate -FilePath $env:CSC_LINK -CertStoreLocation Cert:\LocalMachine\Root -Password (ConvertTo-SecureString -String $env:WIN_CSC_KEY_PASSWORD -AsPlainText -Force)
 
                 # $env:CSC_KEY_PASSWORD=$env:WIN_CSC_KEY_PASSWORD
-                make package
+                make package ELECTRON_BUILDER_ARGS='-c.win.certificateSubjectName="Automattic, Inc."'
       - run:
           name: Archive win-unpacked directories
           command: |
--- a/Makefile
+++ b/Makefile
@@ -209,10 +209,10 @@ docker-build:
 # !! Ensure this file is removed regardless of success/failure !!
 # Note: Ideally we could use a build argument for this, but Docker CE on Windows
 # has trouble consuming the SSH key contents when passed as a build arg.
-	function cleanUp {
+	function cleanup {
 		rm "$(THIS_DIR)/id_rsa"
 	}
-	trap cleanUp EXIT
+	trap cleanup EXIT
 
 	cp "$(SSH_PRIVATE_KEY_FILE)" "$(THIS_DIR)/id_rsa"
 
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -334,7 +334,7 @@ jobs:
                 Import-PfxCertificate -FilePath $env:CSC_LINK -CertStoreLocation Cert:\LocalMachine\Root -Password (ConvertTo-SecureString -String $env:WIN_CSC_KEY_PASSWORD -AsPlainText -Force)
 
                 # $env:CSC_KEY_PASSWORD=$env:WIN_CSC_KEY_PASSWORD
-                make package
+                make package ELECTRON_BUILDER_ARGS='-c.win.certificateSubjectName="Automattic, Inc."'
```
nsakaimbo, line 337: Note: This parameter is required in Circle, but will fail the build locally (unless the user installs the certificate to their certificate store). Therefore we remove it from package.json and pass it only in the Circle job, so packaging the application locally can work as expected (i.e. with the `CSC_LINK/PASSWORD` environment variables).

Makefile
```diff
@@ -66,7 +66,7 @@ dev-server: checks
 	@npx concurrently -k \
 	-n "Calypso,Desktop" \
 	"$(MAKE) calypso-dev NODE_ENV=$(NODE_ENV) CALYPSO_ENV=$(CALYPSO_ENV)" \
-	"wait-on http://localhost:3000 && $(MAKE) build-desktop NODE_ENV=$(NODE_ENV)" \
+	"wait-on http://localhost:3000 && $(MAKE) build-desktop-source NODE_ENV=$(NODE_ENV)" \
```
nsakaimbo, line 69: Note: Decouple building JavaScript source from the native binaries to enable `make dev-server` on Windows. Native binaries are re-built automatically in a few other places (such as npm post-install script and with `make package`), so this change is fine.

PR #830, opened 2020-04-03
style: use title case for menu items

nsakaimbo:
### Description

Per @SylvesterWilmott's recommendation, use Title Case for all menu items.

Performed an audit of the app and this appeared to be the only straggler.


SylvesterWilmott:
Thanks @nsakaimbo !

PR #831, opened 2020-04-03
chore: rename built artifacts

nsakaimbo:
### Description:

There's a few manual steps in the release process that involve renaming built artifacts prior to uploading via Phabricator. This PR amends the build artifact names to match those used uploaded to Phabricator so manual renaming isn't necessary.

Amending the artifact names inline in the package.json config ensures that each platform's `latest.yml` also reflects the correct URL for auto-update. (We will test this with the upcoming beta release and can always revert if this isn't the case.)

<p align="center">
rename-artifacts
</p>

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,8 @@
       "hardenedRuntime": true,
       "extendInfo": {
         "LSFileQuarantineEnabled": true
-      }
+      },
+      "artifactName": "wordpress.com-macOS-app-${version}.${ext}"
     },
     "dmg": {
       "title": "WordPress.com Installer",
@@ -135,7 +136,8 @@
           "y": 179,
           "type": "file"
         }
-      ]
+      ],
+      "artifactName": "wordpress.com-macOS-dmg-${version}.${ext}"
     },
     "win": {
       "target": {
@@ -147,7 +149,8 @@
       }
     },
     "nsis": {
-      "oneClick": false
+      "oneClick": false,
+      "artifactName": "wordpress.com-win32-setup-${version}.${ext}"
     },
     "linux": {
       "target": [
@@ -160,7 +163,8 @@
       "category": "Development",
       "desktop": {
         "StartupNotify": true
-      }
+      },
+      "artifactName": "wordpress.com-linux-x64-${version}.${ext}"
     },
     "deb": {
       "fpm": [
@@ -172,7 +176,8 @@
         "libnotify4",
         "libxtst6",
         "libnss3"
-      ]
+      ],
+      "artifactName": "wordpress.com-linux-deb-${version}.${ext}"
     },
     "afterSign": "./after_sign_hook.js"
   }


nsakaimbo:
@loremattei  We can keep the artifact names consistent with what they've been historically (as in this PR). But it seems there might be an opportunity to make a few filenames more concise? Perhaps:

|   | Before  | After  | |
|---|---|---|---|
|  Win | wordpress.com-win32-setup-$CURRENT_VERSION.exe  | wordpress.com-win32-setup-$CURRENT_VERSION.exe  | keep same |
|  Mac | wordpress.com-macOS-app-$CURRENT_VERSION.zip  | wordpress.com-$CURRENT_VERSION-mac.zip  |
| Mac  | wordpress.com-macOS-dmg-$CURRENT_VERSION.dmg  | wordpress.com-$CURRENT_VERSION.dmg  | |
|  Linux | wordpress.com-linux-x64-$CURRENT_VERSION.tar.gz  | wordpress.com-linux-x64-$CURRENT_VERSION.tar.gz  | keep same |
| Linux  |  wordpress.com-linux-deb-$CURRENT_VERSION.deb  |  wordpress.com-linux-$CURRENT_VERSION.deb   | |

While we're still using the Phabricator workflow, maybe we should keep the names as-is for now but transition to more concise names when we move to Github (thinking of those arrays that need updating - might be nice to be consistent for the time being). 🤔 Would like to change the names eventually. What do you think?

belcherj:
This looks good but I like the idea of using this as an opportunity to update naming conventions.

loremattei:
I don't mind about the artifacts names :-)
But yes, keeping them as-is while we're using the current workflow will make things easier.

loremattei:
I'm trying to test this PR, but I get this error when I start `make dev`:
```
Error: Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@babel/plugin-proposal-nullish-coalescing-operator' from '/Users/lore/Sources/Automattic/wp-desktop'
    at Function.module.exports [as sync] (/Users/lore/Sources/Automattic/wp-desktop/node_modules/resolve/lib/sync.js:69:15)
    at resolveStandardizedName (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/files/plugins.js:101:31)
    at resolvePlugin (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/files/plugins.js:54:10)
    at loadPlugin (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/files/plugins.js:62:20)
    at createDescriptor (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-descriptors.js:154:9)
    at /Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-descriptors.js:109:50
    at Array.map (<anonymous>)
    at createDescriptors (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-descriptors.js:109:29)
    at createPluginDescriptors (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-descriptors.js:105:10)
    at plugins (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-descriptors.js:40:19)
    at mergeChainOpts (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-chain.js:314:26)
    at /Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-chain.js:278:7
    at buildRootChain (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/config-chain.js:89:20)
    at loadPrivatePartialConfig (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/partial.js:85:55)
    at loadFullConfig (/Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/config/full.js:43:39)
    at /Users/lore/Sources/Automattic/wp-desktop/node_modules/@babel/core/lib/transform.js:28:33
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
    at eval (webpack:///./desktop/index.js?:1:7)
    at Object../desktop/index.js (/Users/lore/Sources/Automattic/wp-desktop/build/desktop.js:97:1)
    at __webpack_require__ (/Users/lore/Sources/Automattic/wp-desktop/build/desktop.js:21:30)
    at /Users/lore/Sources/Automattic/wp-desktop/build/desktop.js:85:18
    at Object.<anonymous> (/Users/lore/Sources/Automattic/wp-desktop/build/desktop.js:88:10)
    at Module._compile (internal/modules/cjs/loader.js:968:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:986:10)
    at Module.load (internal/modules/cjs/loader.js:816:32)
    at Module._load (internal/modules/cjs/loader.js:728:14)
    at Module._load (electron/js2c/asar.js:717:26)
```

@nsakaimbo do you have any idea about this?

nsakaimbo:
I just attempted to `make dev` with this branch and it seems to be working for me.

The error message is a little mysterious, but seems to indicate that it's somehow having trouble resolving node module dependency `@babel/plugin-proposal-nullish-coalescing-operator`. 

What happens if you install that dependency explicitly? `i.e. npm install --save-dev @babel/plugin-proposal-nullish-coalescing-operator@7.4.4`

loremattei:
> What happens if you install that dependency explicitly?

Installation succeeds. And after that, `make dev` works. Which makes me wonder why `npm ci` didn't show any failure 🤷‍♂️

Thanks for the suggestion!

loremattei:
I tested this way to verify that the previous versions can download the new binaries:
- set up a fake release in my forked repository with the artifacts from this PR;
- enabled the development more on electron-updater to make it point to that repository. 
- run the dev build of the `develop` branch (which has the older electron-updater configuration).

I confirm that it was able to find and download the new .zip on MacOS. 

This looks good to me. I'll let you decide if you want to test on the other platforms as well before merging (I think it should be good, out of bugs in `electron-updater` itself).

belcherj:
Nice work @nsakaimbo

nsakaimbo:
Thanks for testing, @loremattei! I think we can reasonably expect the auto-updater to work on all platforms, and we can verify when the new beta is deployed. (If not, we can always revert this.)

PR #848, opened 2020-04-17
build: automate Github release notes

nsakaimbo:
### Description

This PR adds a script that gathers changes since the last stable release of the application and formats them using markdown. This markdown-formatted changelog is published with every release to provide a baseline, readily-available view of changes.

In addition, the script is designed such that it can also be executed locally on any branch, and will output the set of changes since the last stable release through the current branch's `HEAD`.

<p align="center">
automated-release-notes-example
</p>

There are a number of frameworks out there that help parse and format changes in the conventional commit format. Inspired by [this blog post](https://nikoheikkila.fi/blog/generating-conventional-changelogs/), I chose to go with a relatively lightweight approach and use a custom bash script instead (that ends up being ~100 LOC). Without the overhead of an entire framework, using bash is also more portable between environments. If we find this concept useful after trialing it out, we can always use a more fully-fleshed framework if we want any additional functionality.

### To Test

* _Release Notes on Publish_: Tag this branch with a temporary, dummy tag (greater than the current stable release version) to create a release draft in Github. Note that the release notes section of the draft will be pre-populated.

* _Using the Script Locally_: Check out this branch and run `./make-changelog.sh` from the root of the repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -435,14 +435,16 @@ jobs:
   publish:
     docker:
       - image: circleci/golang:1.12
+    working_directory: /home/circleci/wp-desktop
     environment:
       VERSION: << pipeline.git.tag >>
     steps:
       - when:
           condition: << pipeline.git.tag >>
           steps:
+            - checkout
             - attach_workspace:
-                at: .
+                at: /home/circleci/wp-desktop
             - run:
                 name: Install Dependencies
                 command: go get github.com/tcnksm/ghr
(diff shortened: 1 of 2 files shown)


designsimply:
This is very cool.

nsakaimbo:
> This is very cool.

Thanks, @designsimply.

We're trying this out to see if it ultimately provides value. The hope is this strategy can provide a baseline summary to stakeholders of what's changed between releases, with little-to-no manual intervention (i.e. a release can be created at any time and will have a changelog in Github). This can also be helpful in focusing efforts when deciding "what to test" when gearing up for a new release.

We'll see how it goes - any and all feedback is welcome!

PR #870, opened 2020-05-01
build: Add mac build step to calypso canary workflow

bsessions85:
### Description:

This change adds the Mac build in to the Calypso canaries workflow. 

### Motivation and Context:
The purpose behind this is that we would like to be able to run the desktop e2e tests on Calypso changes so we can have more confidence that it works

### How Has This Been Tested:

Here is a test build I triggered from the gh bridge https://circleci.com/workflow-run/01bd7a82-b85b-4675-ac61-c60ab74a9000

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,7 +183,7 @@ references:
               "payload": {
                 "outcome": "'"success"'",
                 "status": "'"success"'",
-                "branch": "'"$BRANCHNAME"'",
+                "branch": "'"$CIRCLE_BRANCH"'",
                 "build_url": "'"$CIRCLE_BUILD_URL"'",
                 "build_parameters": {
                   "build_num": '"$CIRCLE_BUILD_NUM"',
@@ -208,7 +208,7 @@ references:
             "payload": {
               "outcome": "'"failed"'",
               "status": "'"failed"'",
-              "branch": "'"$BRANCHNAME"'",
+              "branch": "'"$CIRCLE_BRANCH"'",
               "build_url": "'"$CIRCLE_BUILD_URL"'",
               "build_parameters": {
                 "build_num": '"$CIRCLE_BUILD_NUM"',
@@ -227,7 +227,6 @@ jobs:
       CONFIG_ENV: release
       MINIFY_JS: true
       NODE_ARGS: --max_old_space_size=3072
-      BRANCHNAME: << pipeline.git.branch >>
       CALYPSO_HASH: << pipeline.parameters.CALYPSO_HASH >>
       sha: << pipeline.parameters.sha >>
       calypsoProject: << pipeline.parameters.calypsoProject >>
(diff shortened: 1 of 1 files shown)


nsakaimbo:
@bsessions85 Thanks for getting this wired up! 🎉 

* This looks good, although I think want to make sure that `webhook_notify_failed` can be fired in two places: (1) if the calypso build step fails, and (2) if the mac build step fails. For example, if calypso fails to build for any reason, the downstream `mac` job won't get fired at all (and therefore CI status will be left hanging because `webhook_notify_failed` won't get called).

* We recently added a "semantic PR title" bot to the repo as we're testing release note automation. The bot expects the format `<change type>(<optional scope>): <description>`. For this PR, I'd suggest simply pre-pending your current title with "build:" `build: Add mac build step to calypso canary workflow`

.circleci/config.yml
```diff
@@ -198,24 +198,24 @@ references:
...
-              "sha": "'"$sha"'",
-              "calypsoProject": "'"$calypsoProject"'"
+            "payload": {
+              "outcome": "'"failed"'",
+              "status": "'"failed"'",
+              "branch": "'"$BRANCHNAME"'",
+              "build_url": "'"$CIRCLE_BUILD_URL"'",
+              "build_parameters": {
+                "build_num": '"$CIRCLE_BUILD_NUM"',
+                "sha": "'"$sha"'",
```
nsakaimbo, line 215: @bsessions85 Do these variables need to be exposed in the `mac` job the way they have been in the `build` job to-date?

https://github.com/Automattic/wp-desktop/blob/develop/.circleci/config.yml#L230-L233

If this isn't quite working yet, it's possible these env vars are not being populated because they haven't been exposed in the same way as prior in the `build` job.

Same would go for the `BRANCHNAME` var.

bsessions85: yes, THANK YOU. I totally missed that.

bsessions85:
I did some testing to force failures in build and mac jobs
build:
https://circleci.com/gh/Automattic/wp-desktop/60040
mac:
https://circleci.com/gh/Automattic/wp-desktop/60103

I also made sure the PR was updated accordingly.

.circleci/config.yml
```diff
@@ -198,24 +198,24 @@ references:
...
-            "branch": "'"$BRANCHNAME"'",
-            "build_url": "'"$CIRCLE_BUILD_URL"'",
-            "build_parameters": {
-              "build_num": '"$CIRCLE_BUILD_NUM"',
-              "sha": "'"$sha"'",
-              "calypsoProject": "'"$calypsoProject"'"
+            "payload": {
+              "outcome": "'"failed"'",
+              "status": "'"failed"'",
+              "branch": "'"$BRANCHNAME"'",
```
nsakaimbo, line 211: @bsessions85 Looks good. Minor nit: do we need to use `BRANCHNAME` any longer, if it's the same as `CIRCLE_BRANCH`? Might as well as stick to one or the other to make it a little easier to grok?

.circleci/config.yml
```diff
@@ -378,6 +377,10 @@ jobs:
       xcode: "10.1.0"
     shell: /bin/bash --login
     working_directory: /Users/distiller/wp-desktop
+    environment:
+      BRANCHNAME: << pipeline.git.branch >>
```
nsakaimbo, line 381: Same as prior comment - if we stick to `CIRCLE_BRANCH`, do we need a separate `BRANCHNAME` var? Seems like we could remove this?

Issue #640, opened 2019-06-26
Cannot access block editor via the app when using either a WordPress.com Atomic or a Jetpack site


Robertght:
We have two situations, yet the block editor isn't working in the app.

1. On a Business plan site on WordPress.com

#### Steps to reproduce
1. Try to change the editor to Classic, if you're on the Block one and then back to the Block editor.

#### What I expected
To be able to use the Block editor

#### What happened instead
The page just refreshes and hides the sidebar.

2. On a self-hosted site(tested with Jurassic ninja)

#### Steps to reproduce
1. Create a new site -> Connect it to Jetpack.
2. Select a post or a page and try to switch to the Block editor
3. The page within the app will sort of freeze and open a new one in the browser
4. Click back to return to the page/post list and try to enter again one of them in order to edit it

#### What I expected
To be able to use the Block editor in the app

#### What happened instead
It will redirect you to the browser in order to edit your content.

#### App Version / OS version
WordPress 4.2.0 / MacOS 10.14.5

#### Screenshot / Video
https://monosnap.com/file/wgDgF02nQuGRN8hKrT9VGt2uTer4MK

#### Context / Source
#user-report #manual-testing

#2135575-zen

rachelmcr:
Thanks for the report! I confirmed that on a WordPress.com Atomic site, the app consistently redirects to the Calypso classic editor, regardless of the block editor setting/switch. The block editor switch works for WordPress.com Simple sites as expected.

We're in a bit of an in-between stage with the desktop app. The block editor in Calypso is supported for Jetpack/Atomic sites on the web, but it isn't yet enabled in the app. You can refer to Automattic/wp-calypso#32048 for progress on enabling that in the desktop app.

The current behavior might be what's expected for Atomic sites at the moment, although I'm not sure. The other behavior you described (with a Jetpack site) is currently the expected behavior for Jetpack sites (since the block editor isn't supported in the app, it opens in a browser tab instead). @apeatling should the behavior for Atomic and Jetpack sites be the same in the app right now (either both forced to use the classic editor, or both opening the block editor in a browser)?

SiobhyB:
This came up again in 2144735-zen. The user has an Atomic site and would like to use the new block editor in the desktop app, but they're not able to load it.

designsimply:
Tested and confirmed that I cannot use the block editor for Atomic or Jetpack sites in WPDesktop 4.3.0-beta1. 

[...]

PR #876, opened 2020-05-08
fix: ensure Jetpack redirects don't block the app

nsakaimbo:
### Description

This PR addresses #640, in which a number of Jetpack users have reported the app being left in an unusable state when trying to use the Gutenberg editor.

<p align="center">

<p>

[Closer investigation](https://github.com/Automattic/wp-desktop/issues/640#issuecomment-621268932) confirmed that because the Desktop app does not support Jetpack authentication, this happens when Calypso remains stuck in a "pending" state while the user is redirected to their remote site for re-authentication. **Workaround**: in order for Gutenberg to work in the Desktop app, the user has to enable Single Sign-On in the remote site's Jetpack settings.

~~This PR intercepts the authentication redirect from Jetpack and lets the user know that SSO is required. The user can then choose to cancel the navigation or proceed in an external browser. With either selection, the user is automatically navigated to the `my-sites` landing page so they are not left in an unusable state.~~

~~This fix is regarded as a stopgap while a more ideal solution is pending (that will likely require support from both Jetpack and Calypso). Ideally, Jetpack users should have a more seamless way/less disruptive way to authenticate within the Desktop app without having to know _in advance_ that SSO should be enabled.~~

<p align="center">
enable-sso-prompt
</p>

This PR intercepts the authentication redirect from Jetpack and allows the user to enable SSO, proceed in a browser or cancel navigation to the editor. When proceeding to an external browser or canceling, the user is automatically navigated to the `my-sites` landing page so they are not left in an unusable state.

This PR pairs with the changes in Automattic/wp-calypso#42183.

### Additional Context: Classic Editor Deprecation

Using the Classic Editor as a fallback was considered but is a no-go as the Classic Editor will imminently be deprecated from Calypso (see [this master Github issue](Automattic/Dotcom-roadmap/issues/124) for progress).

### To Test

- Download and extract the [mac artifact built with this branch](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/14504/workflows/0c5ce06d-acf9-4fbf-9bed-7f3011be3ae4/jobs/62610/artifacts) (use `ditto -x -k wordpress.com-macOS-app-5.1.1-beta1.zip .` to extract. Double-click to open and then select "Open Anyway" in MacOS `Security & Privacy` preferences since the app not notarized)
- Set up a self-hosted site (with Jurassic Ninja) and add a WordPress.com user. Connect and activate Jetpack!
- Log into the desktop app as the WordPress.com user 

With Gutenberg enabled:

- Attempt to use the editor in a simple site: expect the editor to work as before
- Attempt to use the editor in a remote site: UI prompt to continue in a browser is displayed. If the user is an admin (i.e. can manage site options), they will also have the option to enable SSO and proceed to the Gutenberg editor.

Related: Automattic/wp-calypso/issues/41417

### DIFF
--- a/desktop/window-handlers/external-links/editor/index.js
+++ b/desktop/window-handlers/external-links/editor/index.js
@@ -1,0 +1,177 @@
+'use strict';
+
+/**
+ * External Dependencies
+ */
+const { URL } = require( 'url' );
+const { promisify } = require( 'util' );
+const { dialog, ipcMain: ipc } = require( 'electron' );
+
+/**
+ * Internal dependencies
+ */
+const Settings = require( 'lib/settings' );
+const openInBrowser = require( '../open-in-browser' );
+const { showMySites } = require( 'lib/calypso-commands' );
+const settingConstants = require( 'lib/settings/constants' );
+const log = require( 'lib/logger' )( 'desktop:external-links:handle-jetpack' );
+
+/**
+ * Module variables
+ */
+const delay = promisify( setTimeout );
+
+// Invokes the activation of Secure Sign-On ("SSO") for the remote site
+// by sending messages over IPC to Calypso. Because we should not block
+// the renderer process by sending mesages synchronously, each step sets
+// and cleans up a handler that listens on a corresponding response channel.
+//
+// Steps:
+// 1. Enable option (set/clean up handler on response channel)
+// 2. Refresh site info (set/clean up handler on response channel)
+// 3. Navigate to the editor
+function selectedEnableSSOandContinue( mainWindow, info ) {
+	log.info( 'User selected \'Enable SSO and Continue\'...' );
+
+	const channelDidEnableSiteOption = 'enable-site-option-response';
+	const channelDidRequestSite = 'request-site-response';
+
+	const { siteId, origin, editorUrl } = info;
+
+	const handleDidRequestSite = ( _, { error } ) => {
+		ipc.removeAllListeners( channelDidRequestSite );
+		if ( error ) {
+			throw error;
+		}
+		log.info( `Refreshed info for site: '${ origin }', navigating to URL: '${ editorUrl }'` );
+		mainWindow.webContents.send( 'navigate', editorUrl );
+	}
+
+	const handleDidEnableSiteOption = ( _, { error } ) => {
+		ipc.removeAllListeners( channelDidEnableSiteOption );
+		if ( error ) {
+			throw error;
+		}
+		log.info( `SSO enabled for site: '${ origin }', requesting site refresh...`, );
+		ipc.on( channelDidRequestSite, handleDidRequestSite );
+		mainWindow.webContents.send( 'request-site', siteId );
+	}
+
+	ipc.on( channelDidEnableSiteOption, handleDidEnableSiteOption );
+	mainWindow.webContents.send( 'enable-site-option', { siteId, option: 'sso' } );
+}
+
+function selectedProceedInBrowser( mainWindow, { origin, wpAdminLoginUrl } ) {
+	log.info( 'User selected \'Proceed in Browser\'...' )
+	openInBrowser( null, wpAdminLoginUrl );
+	navigateToShowMySites( mainWindow, origin );
+}
+
+function selectedCancel( mainWindow, { origin } ) {
+	log.info( 'User selected \'Cancel\' with origin ...', origin );
+	navigateToShowMySites( mainWindow, origin );
+}
+
+function navigateToShowMySites( mainWindow, domain ) {
+	showMySites( mainWindow );
+	// Update last location so a redirect isn't automatically triggered on app relaunch.
+	Settings.saveSetting( settingConstants.LAST_LOCATION, `/stats/day/${ ( new URL( domain ) ).hostname }` );
+}
+
+/**
+ * Prompts the user when they are unable to navigate to the Gutenberg editor due to lack
+ * of sufficient permissions for the self-hosted site.
+ *
+ * The user may select one of three options to continue:
+ *
+ *	1. Enable Secure Sign On and Continue
+ *	2. Proceed in Browser (via wpAdmin redirect)
+ *	3. Cancel
+ * Users that can manage site options will be presented with all three options. All others
+ * are presented with the option to Cancel or Proceed in Browser only.
+ *
+ * FIXME: Implementing this logic (including UI dialogs) may be a lot cleaner if done
+ * wholly within Calypso (minimize necessity of async workarounds over IPC).
+ *
+ * @param {object} mainWindow Reference to the main window object.
+ * @param {object} info Payload from Calypso containing relevant user- and site-specific information.
+ */
+async function handleJetpackEnableSSO( mainWindow, info ) {
+	log.info( `Prompting user to enable SSO for site: '${ info.origin }'...` );
+
+	const { origin, canUserManageOptions } = info;
+
+	const buttons = [ 'Proceed in Browser', 'Cancel' ];
+	if ( canUserManageOptions ) {
+		buttons.unshift( 'Enable SSO and Continue' );
+	}
+
+	try {
+		// Allow sufficient time for Gutenberg's "placeholder" UI to render.
+		// Otherwise, exiting the placeholder UI will make Calypso's master- and
+		// sidebars disappear.
+		await delay( 300 );
+
+		let message = 'This feature requires that Secure Sign-On is enabled in the Jetpack settings of the site:' +
+					'\n\n' +
+					`${ origin }`;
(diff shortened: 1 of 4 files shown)


desktop/window-handlers/external-links/index.js
```diff
@@ -45,8 +49,16 @@ function isValidBrowserUrl( url ) {
 	return false;
 }
 
+function isWpAdminLogin( url ) {
```
nsakaimbo, line 52: Ideally, we should be able to get a site's definitive Jetpack designation, but I'm unsure how to do this. This is the best proxy I could come up with by inspecting the Jetpack redirect URLs.

If you use the WP.com API to query a site with `GET https://public-api.wordpress.com/rest/v1.1/<domain>`, one of the elements returned in the JSON response is a `jetpack` boolean. This is even better, but we would have to insert a blocking HTTP request to fetch this info - not sure if that's a good strategy, but is certainly an option.

It's possible we might get this from Calypso for free somehow, but am unsure where to look/how to hook into this info.

dmsnell: this should be available in the [`getSelectedSite()`](https://github.com/Automattic/wp-calypso/blob/master/client/state/ui/selectors/get-selected-site.js) selector. You'll need to get it from the Redux store but I think you can do that with a single import - it could be more than this but regardless it's possible.

after that you should even have access to [`getJetpackSetting()`](https://github.com/Automattic/wp-calypso/blob/master/client/state/selectors/get-jetpack-setting.js) or whatever it is that will tell us if we have SSO enabled.

pento:
I've just been doing some testing against pento.wpsandbox.me, unfortunately I haven't been able to get it to load the block editor, it only shows me the pop-up. SSO is definitely enabled, and I've confirmed that it loads correctly in Calypso in my browser. Is there any easy way for me to get you debug information, or is it easiest if I just build wp-desktop locally?

A few other notes I had:

- Jetpack's SSO actually stands for "[Secure Sign On](https://jetpack.com/support/sso/)", rather than Single Sign-On.
- Need to remove the `calypsoify=1` parameter from the URL opened in the browser, as I suspect folks will be surprised about their `wp-admin` view suddenly being blue. 🙂
- It would be neat if we could enable SSO from the app, rather than making the end users guess what they need to do. It seems the UI exists (Manage->Settings->Security), can we direct them to there?
- The pop-up is really information dense: there's a lot of jargon that end users won't (and shouldn't have to) understand.

Once we figure out why I'm getting the pop-up even with SSO enabled, I'm happy to test what happens when they Jetpack site has the Classic Editor plugin enabled, too. 🙂

nsakaimbo:
Thanks for testing, @pento!

I'm marking this PR as "In Progress" as there's clearly more work to do.

I've started to look at making changes within Calypso, so Calypso can tell the Desktop app when a user is unable to use the Block Editor. By letting Calypso do the heavy lifting we can avoid the Desktop having to make "educated guesses" based on the responses/redirects from Jetpack, which is very messy and prone to error. This will also avoid duplication of logic since Calypso has all the site-specific information it needs. Hopefully this will also avoid some of the inconsistencies you're experiencing when testing as well.

> the pop-up is really information dense: there's a lot of jargon that end users won't (and shouldn't have to) understand.

Agreed. As I mentioned in the PR description, this is a first pass at unblocking users, because at the moment this flow puts them in an unusable state. The ideal would be _not_ to have their experience interrupted with a pop-up or jargon, but this might be unavoidable in the interim. If you have alternate language you think would be best, I'm happy to change it. Even if a redirect to their Site/Plugin Settings is possible, we'll need some sort of way to tell them _what_ to change. 

We'll continue to iterate and hopefully give users a better experience, but at the moment the priority is to stop the bleeding.

> I'm happy to test what happens when they Jetpack site has the Classic Editor plugin enabled, too

Depending on the additional effort, that may/may not be part of a separate PR/effort. Even though related, we shouldn't block one fix that's ready to go with another. From our users' perspective, this is a very (very!) bad bug, and I'd like to roll out whatever fixes we can as they become ready.

desktop/window-handlers/external-links/index.js
```diff
@@ -103,7 +124,46 @@ module.exports = function( webContents ) {
...
+				}
+				showMySites( mainWindow );
+				// Update last location so a redirect isn't automatically triggered on app relaunch
+				Settings.saveSetting( settingConstants.LAST_LOCATION, `/stats/day/${parsedURL.hostname}` );
+			} catch ( error ) {
+				log.error( 'Failed to prompt for Jetpack authorization: ', error );
+			}
+		}
+
+		await delay( 300 );
```
dmsnell, line 165: can we leave an explanatory comment on why we're waiting and why we're waiting 300ms?

 - Is there an underlying race condition we're trying to avoid?
 - are we trying to let network requests settle?
 - are we only trying to avoid instantly popping up the window?

desktop/window-handlers/external-links/index.js
```diff
@@ -103,7 +124,46 @@ module.exports = function( webContents ) {
...
-		log.info( `Using system default handler for URL: ${ openUrl }` );
 		openInBrowser( event, openUrl );
 	} );
+
+	ipc.on( 'cannot-open-editor', async( event, info ) => {
+		const { origin, editorUrl } = info;
+		const parsedURL = new URL( origin );
+
+		const promptWpAdminAuth = () => {
+			try {
```
dmsnell, line 135: what's the source of the exception we're catching?

nsakaimbo: I think being overly paranoid about showMessageBox failing, but you're right - this is probably unnecessary.

desktop/window-handlers/external-links/index.js
nsakaimbo: Will leave a note. 

It works just fine without the delay, except if one of the buttons presented _without_ the [Gutenberg placeholder UI](https://github.com/Automattic/wp-calypso/blob/master/client/gutenberg/editor/controller.js#L110-L112) having a chance to display, dismissing the modal will leave the UI in a state where the Calypso "chrome" (masterbar, etc) disappears from the current view. It's really odd - I should take a video to explain what I mean. 

This just gives the placeholder UI a chance to display, so that things look as expected when dismissed.

desktop/window-handlers/external-links/editor/index.js
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * @param {object} mainWindow Reference to the main window object.
+ * @param {object} info Payload from Calypso containing relevant user- and site-specific information.
+ */
+async function handleJetpackEnableSSO( mainWindow, info ) {
+	log.info( `Prompting user to enable SSO for site: '${ info.origin }'...` );
+
+	const { origin, canUserManageOptions } = info;
+
+	let buttons = [ 'Proceed in Browser', 'Cancel' ];
```
belcherj, line 104: Make const

desktop/window-handlers/external-links/editor/index.js
```diff
@@ -0,0 +1,177 @@
...
+	if ( canUserManageOptions ) {
+		buttons = [ 'Enable SSO and Continue', 'Proceed in Browser', 'Cancel' ];
```
belcherj, line 106: buttons.unshift('Enable SSO and Continue')

public_desktop/preload.js
```diff
@@ -27,8 +27,6 @@ window.logger = logger; // expose logger interface to other renderer processes
 const log = logger( 'desktop:preload' );
 const desktop = remote.getGlobal( 'desktop' );
 
-log.debug( 'Setting up preload script' );
```
belcherj, line 30: maybe separate PR?

nsakaimbo: Reverted this change (unnecessary).

aaronrobertshaw:
> Download and extract the mac artifact built with this branch (use ditto -x -k wordpress.com-macOS-app-5.1.1-beta1.zip . to extract. Double-click to open and then select "Open Anyway" in MacOS Security & Privacy preferences since the app not notarized)

Using this beta release I do not get the prompt at all. I just get redirected to the browser with no prompt. 

If I checkout the PR branch and rebuild the app locally, the prompt displays and tests ok for me.

nsakaimbo:
@aaronrobertshaw 

> Using this beta release I do not get the prompt at all. I just get redirected to the browser with no prompt.

Thanks for helping test! You would need to download the app from the artifacts link for this particular PR - this build isn't available as an official release yet. Here's the [link](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/14504/workflows/0c5ce06d-acf9-4fbf-9bed-7f3011be3ae4/jobs/62610/artifacts). Will clarify in the PR comments as well.

Glad you were able to confirm things were working with a local build of this branch. 🎉

PR #878, opened 2020-05-12
revert(os x): com.apple.quarantine attr

nsakaimbo:
### Description

This PR reverts the addition of the `com.apple.quarantine` attribute (introduced in [this](https://github.com/Automattic/wp-desktop/pull/738) PR), which can have unintended side effects w.r.t. the Electron's auto-updater.

TL;DR: the `com.apple.quarantine` attribute can make it so the auto-updater is unable to successfully update the app. This manifests as the error `Update error: Cannot update while running on a read-only volume.` in the Electron Updater logs. (Note: this doesn't occur 100% of the time, but does seem to be an unresolved issue that occurs intermittently).

Github Issue: https://github.com/Squirrel/Squirrel.Mac/issues/182

Manual Workarounds:

- Option 1: Remove `com.apple.quarantine` attribute prior to updating: `xattr -dr com.apple.quarantine `
- Option 2: Using Finder, move the application from the /Applications directory to another location and back

While a more robust workaround is pending, will revert the addition of this attribute.

Note: This fix is forward-looking - the auto-update code lives within the app that's currently installed, not the app being deployed, so we can't override this attribute retroactively. This fix will help mitigate the issue for users of the app going forward (i.e. those users that install >=5.1.1), but we'll have to let Happiness know about the manual workarounds for users that may encounter it in the interim.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,8 +25,8 @@ references:
     restore_cache:
       name: Restoring NVM cache
       keys:
-        - v1-nvm-0-33-11-{{ arch }}-{{ checksum "calypso/.nvmrc" }}
-        - v1-nvm-0-33-11-{{ arch }}
+        - v1-nvm-0-33-11-{{ .Environment.CACHE_VERSION }}-{{ arch }}-{{ checksum "calypso/.nvmrc" }}
+        - v1-nvm-0-33-11-{{ .Environment.CACHE_VERSION }}-{{ arch }}
   setup_nvm: &setup_nvm
     run:
       name: Install nvm and calypso node version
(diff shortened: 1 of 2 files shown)


nsakaimbo:
**Update**: This PR now contains reference to a cache reset keys that allow us to continue building the desktop app.

Unfortunately a build-time blocker occurred over the last couple of days that retroactively corrupted the built, cached Calypso source (see [this issue](https://github.com/Automattic/wp-calypso/issues/42163)). The updated cache keys allow us to invalidate the corrupt cache and re-build this specific SHA of Calypso with an _unminified_ config, which is seemingly our only path forward at the moment.

Smoke-testing indicates the application bundle size is still more than reasonable (built JavaScript source file `build/desktop.js` is ~6MB), and there's no noticeable impact to the performance/responsiveness of the app. Pre-release testing of the 5.1.1 release will have to be extended in order to ensure that unexpected side effects as a result of the change.

PR #884, opened 2020-05-19
chore: update babel deps

nsakaimbo:
### Description

Updating babel deps to address a build/runtime error that was introduced in Automattic/wp-calypso@0d89b95c8980d15ec084635f1e4e145ecfefff6b.

```
ERROR in ./calypso/client/state/login/actions/reboot-after-login.ts
Module build failed (from ./node_modules/babel-loader/lib/index.js):
SyntaxError: /home/circleci/wp-desktop/calypso/client/state/login/actions/reboot-after-login.ts: Unexpected token (4:12)
```

Hopefully this fixes the above. 🤞

_Update_

Updating babel led to another runtime error: `Cannot assign to read-only property 'exports' of object ...`

* Updating babel (and webpack) wasn't sufficient to address this error. 
* I also tried setting [`esModules: true`](https://github.com/Automattic/wp-desktop/pull/884/commits/c21f341cfa6a120e515201ce45b8dbd4dff03851) in tandem with [`modules: false`](https://github.com/Automattic/wp-desktop/pull/884/commits/23c77f164cc3d998bd064a153e32af778148a341), and that didn't seem to help either. 
* What does seem to have worked is integrating the [babel-plugin-transform-es2015-modules-commonjs](https://github.com/Automattic/wp-desktop/pull/884/commits/a0204112ab5b092a578de4e69000705a055ad66f). From the module's [description](https://www.npmjs.com/package/babel-plugin-transform-es2015-modules-commonjs) page:

> Babel 6 changed some behavior by not doing module.exports = exports['default'] anymore in the modules transforms.
> 
> There are some caveats, but you can use babel-plugin-add-module-exports, so that updating to Babel 6 isn't a breaking change since users that don't use ES modules don't have to do require("your-module").default.
> 
> However, it may not match how Node eventually implements ES modules natively given the the current proposal.

[A comment from a Github issue](https://github.com/webpack/webpack/issues/4039#issuecomment-274240819) also states:

> Use transform-es2015-modules-commonjs. This would replace all import statements with require statements. If you are not using any import/export syntax in your app, this solution is probably ideal, because it incurs no extra cost. However, if you are using ES modules, then this will likely impact Webpack 2's ability to perform tree shaking. (I'm sure everyone is aware of this.)

Given the above, are we in fact losing tree shaking with the desktop build? Not sure what the tradeoffs are here. Any thoughts, @sgomes @belcherj ?


nsakaimbo:
**Update**: Runtime error: `Cannot assign to read-only property 'exports' of object ...`

nsakaimbo:
@belcherj pointed out that the root cause of this issue appears to be the `import type` syntax introduced [here](https://github.com/Automattic/wp-calypso/commit/0d89b95c8980d15ec084635f1e4e145ecfefff6b#diff-895e5d614176bc30ec6850de7b0c37e8R4).

Based on updates to [this](https://github.com/babel/babel/issues/10981) issue, doesn't appear to be addressed in Babel yet.

Is there a smaller change we can make to the desktop webpack config, or is this plugin necessary?

sgomes:
> Based on updates to [this](https://github.com/babel/babel/issues/10981) issue, doesn't appear to be addressed in Babel yet.

That's incorrect; recent versions of `@babel/plugin-transform-typescript` support this feature.

sgomes:
To clarify, there are two separate issues here.

The original issue is that Calypso now has code that relies on TypeScript 3.8. The Desktop app is on an older version of Babel, so in order to support this, a Babel upgrade is needed.

The second issue comes from the Babel update. Something has changed in the way the interaction with ES modules takes place, and the Desktop app's build configurations will need to change as a result. That's the cause of the runtime error.

Hope this helps!

jsnajdr:
> Updating babel led to another runtime error: `Cannot assign to read-only property 'exports' of object ...`

This error happens when you mix ESM and CommonJS syntax inside one module:
```js
import React from 'react'; // ESM import

module.exports = function() { ... } // CJS export
```

When webpack detects that it's dealing with an ES module (detection looks for at least one `import` or `export` statement), it will create a `module` object that has `exports` property protected against writing.

Calypso's `server/pages/index.js` module is exactly this kind of hybrid module. It's a bug that should be fixed.

**Why it normally works?**
In Calypso, the Babel config [specifies `modules: 'commonjs'`](https://github.com/Automattic/wp-calypso/blob/master/babel.config.js#L5) for the `env` preset when building the Node.js server. Then Babel transpiles the ESM syntax into CommonJS, and webpack will only ever see CommonJS modules.

In `wp-desktop`, `modules` is not specified and defaults to `auto`. That means `commonjs` with some caveats. The code that calls Babel, in our case it's `babel-loader`, can specify if the target environment supports ESM and `auto` will be set according to that.

This PR upgraded `babel-loader` to 8.1.0. This version has an [interesting new feature:](https://github.com/babel/babel-loader/releases/tag/v8.1.0)
```
feat: expose webpack target via babel caller
```
Ha! Until now, `modules: 'auto'` defaulted to `commonjs`, but now the `babel-loader` started telling Babel that the caller (webpack) supports ESM! Suddenly modules are no longer transpiled.

> I also tried setting `esModules: true` in tandem with `modules: false`

The right fix is `modules: 'commonjs'` 🙂 The `babel-plugin-transform-es2015-modules-commonjs` plugin is a legacy thing created for Babel 6. But Babel 7 has this functionality built-in. I'll spin up a PR with a change.

sgomes:
Amazing sleuthing work, @jsnajdr! 👍

nsakaimbo:
Fantastic work identifying the root cause, @jsnajdr. And thanks for the overview regarding how Babel and webpack work together w.r.t modules - was unaware and definitely learned something new. PR #887 looks good to me - thanks for getting this fixed!

jsnajdr:
> thanks for the overview regarding how Babel and webpack work together w.r.t modules

The part where `babel-loader` passes down info about the calling environment -- I didn't know about that either. Learned about it for the first time when investigating why the module transpilation disappeared 🙂

PR #887, opened 2020-05-25
fix: force Babel to transpile ESM imports to CommonJS

jsnajdr:
Followup to #884 that improves the `commonjs` fix. Don't use the legacy Babel 6 plugin, but use `preset-env` built-in functionality by specifying `modules: 'commonjs'`.

Previously, the `modules` option defaulted to `auto`, and got broken by a recent `babel-loader` update.

See https://github.com/Automattic/wp-desktop/pull/884#issuecomment-633522275 for more technical details.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -4,7 +4,12 @@ module.exports = {
 	presets: [
 		[
 			'@babel/env',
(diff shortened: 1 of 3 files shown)


nsakaimbo:
LGTM - please merge! And thanks for the additional context added to #884. 👍

PR #890, opened 2020-05-28
build: convert desktop/server/server module to ESM for better Calypso import compat

jsnajdr:
Converts the `desktop/server/server` module from CommonJS to ESM, with `import` and `export` statements.

The benefit is that this creates compatibility with Calypso https://github.com/Automattic/wp-calypso/pull/42643 PR, which converts `calypso/client/server/boot` to ESM, too. Importing that module with
```js
import boot from 'server/boot';
```
works equally well both for ESM and CJS, while
```js
const boot = require( 'server/boot' );
```
works only if `server/boot` is CJS. An ESM module would need
```js
const boot = require( 'server/boot' ).default;
```
Therefore, this PR makes `wp-desktop` compatible with Calypso both pre- and post- https://github.com/Automattic/wp-calypso/pull/42643

I also removed the `use strict` statement, as webpack adds it automatically to all processed modules.

### DIFF
--- a/desktop/server/server.js
+++ b/desktop/server/server.js
@@ -4,12 +4,12 @@
 import { dialog } from 'electron';
 import http from 'http';
 import portscanner from 'portscanner';
-import logFactory from 'lib/logger';
 
 /**
  * Internal dependencies
  */
 import Config from 'lib/config';
+import logFactory from 'lib/logger';
 import boot from 'server/boot';
 
 const log = logFactory( 'desktop:server' );
--- a/desktop/server/server.js
+++ b/desktop/server/server.js
@@ -1,59 +1,62 @@
-'use strict';
-
 /**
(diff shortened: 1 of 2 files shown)


desktop/server/server.js
```diff
@@ -1,59 +1,62 @@
...
-
 /**
  * External Dependencies
  */
-const portscanner = require( 'portscanner' );
-const log = require( 'lib/logger' )( 'desktop:server' );
+import { dialog } from 'electron';
+import http from 'http';
+import portscanner from 'portscanner';
+import logFactory from 'lib/logger';
```
sgomes, line 7: Looks like this should be moved to the internal dependencies section.

jsnajdr: Done! :+1:

PR #896, opened 2020-06-02
build(electron): upgrade to Electron 8 LTS

nsakaimbo:
### Description

This PR updates the application from Electron v4 to Electron v8.

### Key Changes:

- In order to share state with spawned `BrowserWindow` objects, Electron now requires that the preload file is explicitly injected into each child window.
- **Spellchecking**: Electron 8 introduced os-native spellchecking, which drops the need for the essentially unmaintained electron-spellchecker lib. (In fact, I couldn't get `electron-spellchecker` to work with Electron 7 or 8, so native spellchecking support is timely). The native spellchecker is surprisingly straightforward to implement ([docs](https://www.electronjs.org/docs/tutorial/spellchecker)). 
- Breaking API changes: surprisingly not many, although there was an undocumented breaking change to the `session.setProxy` API (replaces callback with a promise).
- **Simplified build**:
  - Workarounds to support electron-spellchecker compilation are no longer necessary (e.g. use of `.npmrc` for headers and `node-gyp` for Windows).

### How Was This Tested

- Smoke-testing (navigation, spellchecking, preferences, shortcuts, etc):
  - [x] Windows
  - [x] Mac
  - [x] Linux

- Confirmed end-to-end tests passing on Mac and Linux.

### Spellchecking and Context Menu

The new spellchecker leans on each operating system's native implementation, and its API is fairly easy to use. For a first pass, the following behavior is implemented.

1. When right-clicking a text element, a context menu will display.

2. If the text is *editable*, show the full menu that includes `Cut`, `Copy`, `Paste` and `Select All` operations. In addition, if spellchecking is enabled, the user can fix a highlighted misspelling with a suggestion, or `Add to Dictionary`.

<p align="center">
context-menu-win
<p>

3. If the text is *non-editable*, the user will only be shown the option to `Copy` the selected text.

<p align="center">
context-menu-partial
<p>

### To Test

Download an [installer artifact](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/15063/workflows/cf58ff60-9231-4e80-b490-a7c78c846ce9/jobs/63301/artifacts) from this PR and give it a whirl!

**Update**: Update artifacts for v5.2.0, to avoid the auto-update prompt ([link](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/15154/workflows/6e612069-1b80-41fe-acc9-b0466b4b6066/jobs/63436/artifacts)).


Makefile
```diff
@@ -160,17 +160,6 @@ else
 	@echo $(GREEN)$(CHECKMARK) Current NodeJS version is on par with Calypso \($(CALYPSO_NODE_VERSION)\) $(RESET)
 endif
 
-# Check that the electron version specified in .npmrc and package.json are consistent.
-# This is to ensure that when native dependencies have an auto-build postinstall script,
-# node-gyp will compile against Electron and not the host environment's node version
-# required by Calypso.
-check-electron-version-parity:
```
nsakaimbo, line 167: Note: this check is no longer needed.

package.json
```diff
@@ -16,7 +16,6 @@
   "main": "build/desktop.js",
   "scripts": {
     "preinstall": "make check-version-parity",
-    "postinstall": "make rebuild-deps",
```
nsakaimbo, line 19: Unnecessary as `electron build` rebuilds native deps anyway when building artifacts. (Also this somehow didn't jive well when trying to update deps via yarn - perhaps because yarn re-installs and compiles native modules in parallel? 🤷 ) 

For non-final/dev builds (e.g. `make dev`), the makefile will explicitly invoke `make rebuild-deps` so we're covered on both fronts.

desktop/lib/window-manager/index.js
```diff
@@ -63,26 +64,35 @@ function setDimensions( config ) {
...
-			Config[settings.config] = setDimensions( Config[settings.config] );
+			let config = Config[settings.config];
+			config = setDimensions( config );
 
-			windows[windowName].handle = new BrowserWindow( Config[settings.config] );
-			windows[windowName].handle.setMenuBarVisibility( false );
-			windows[windowName].handle.webContents.session.setProxy( { proxyRules: 'direct://' }, function() {
-				windows[windowName].handle.loadURL( Config.server_url + ':' + Config.server_port + '/desktop/' + settings.file );
+			// nodeIntegration is necessary to support usage of `require` in preload and HTML scripts.
+			// FIXME: Refactor preload to use contextBridge instead (contextIsolation: true, nodeIntegration: false).
```
nsakaimbo, line 76: An earlier draft of this upgrade went down the route of using `contextBridge`. However I wanted to keep the diff of this PR somewhat reasonable, and that work will be part of a downstream PR.

desktop/app-handlers/preferences/index.js
```diff
@@ -15,7 +15,7 @@ const Settings = require( 'lib/settings' );
 function promptForRestart( title, message ) {
 	// Warn user they need to restart the app
 	dialog.showMessageBox( {
-		buttons: [ 'Ok' ],
```
nsakaimbo, line 18: This hurt my eyes.

Side note: The `spellcheck` property on the window's web preferences cannot be overriden _after_ a `BrowserWindow` is instantiated, so we still need this prompt to restart the app for the setting to take effect.

desktop/window-handlers/spellcheck/index.js
```diff
@@ -0,0 +1,55 @@
...
+				} ) );
+			}
+
+			// Allow users to add the misspelled word to the dictionary
+			if ( params.misspelledWord ) {
+				menu.append( new MenuItem( { type: 'separator' } ) );
+				menu.append(
+					new MenuItem( {
+						label: 'Add to Dictionary',
+						click: () => mainWindow.webContents.session.addWordToSpellCheckerDictionary( params.misspelledWord ),
```
dmsnell, line 32: does this persist across app runs?

nsakaimbo: @dmsnell Yep - it does. This will add suggestions to the user's system dictionary, so the changes will be available from that point on. Words added to the dictionary will be available from that point on (and in fact be available to other applications once added).

nsakaimbo:
Thanks for the review, @dmsnell! re: the dictionary - responded to your question (TLDR it adds the selected word to the user's system dictionary). For the most part, I lifted the spellchecker code directly from Electron's own [docs](https://www.electronjs.org/docs/tutorial/spellchecker), and I've been really impressed with how consistent and robust it is across platforms.

Overall I'm happy with the fact that we're able to upgrade the repo with relatively reasonable diff. Was just looking for some feedback on anything that jumped out.

Feel free to give any of the installers a spin to smoke test, otherwise would appreciate a 👍 if all looks reasonable.

.circleci/config.yml
```diff
@@ -330,12 +330,6 @@ jobs:
       - run:
           name: Install Node Modules
           command: |
-                # The release of nodejs currently being used by Calypso includes a buggy version
```
belcherj, line 333: Was this resolved?

.npmrc
```diff
@@ -1,7 +0,0 @@
-engine-strict = true
```
belcherj, line 1: Why do we no longer need this?

Makefile
belcherj: Why do we no longer need this?

desktop/lib/window-manager/index.js
belcherj: It is for sure safer to have it off but the assets are bundled so there should not be an issue here.

desktop/window-handlers/spellcheck/index.js
belcherj: This is nice! This does not work with electron spellchecker.

belcherj:
Enabled spellchecker. Said I needed to restart. Was in Gutenberg editor. Closed app. Started it again. App crashed. Opened again. Given log in screen.

.circleci/config.yml
nsakaimbo: We're no longer compiling `electron-spellchecker`, so this workaround for node-gyp isn't necessary. We can always revive it if necessary. But for now it's additional complexity we don't need.

nsakaimbo:
> Enabled spellchecker. Said I needed to restart. Was in Gutenberg editor. Closed app. Started it again. App crashed. Opened again. Given log in screen.

@belcherj I looked at your logs and this happened because you accepted the prompt to auto-update the app (this app is on v5.1.1, and you attempted to update to v5.2.0). The app crashed because the app isn't notarized. Also, I believe you can't update the app when it's in a location other than the `/Applications` folder.

```
[2020-06-03 12:23:10.622] [desktop:updater:auto] [error] Update error: Could not get code signature for running application{}
Error: Could not get code signature for running application
    at Server.<anonymous> (/Applications/WordPress.com.app/Contents/Resources/app/build/desktop.js:95:400108)
    at Object.onceWrapper (events.js:299:28)
    at Server.emit (events.js:210:5)
    at emitListeningNT (net.js:1334:10)
    at processTicksAndRejections (internal/process/task_queues.js:79:21)
```

I'll go ahead and rebase on 5.2.0 so the auto-update prompt is disabled.

nsakaimbo:
@belcherj 

Same as with the `node-gyp` workaround, the `electron-version-parity` is no longer needed as we're dropping `electron-spellchecker`. Likewise, the entirety of the `.npmrc` file was to ensure that electron-spellchecker compiled against the correct version of Electron, as `node-gyp` and `electron-builder` didn't seem to handle rebuilding that particular dependency smartly enough by themselves.

`electron-spellchecker` had a number of quirks in its build process, namely post-install build scripts in its sub-dependencies that would attempt to compile against the host Node environment instead of Electron (which would result in a build failure, as `electron-spellchecker` was built with Electron as its target). The use of the .`npmrc` was to force those sub-dependencies to correctly compile against Electron instead.

PR #898, opened 2020-06-04
build(ci): isolate canary from wp-desktop workflows

nsakaimbo:
### Description

This PR attempts to isolate the canary from wp-desktop workflows in Circle as we observed the desktop e2e suite not getting triggered with a number of Calypso PRs. Pairs with recent changes made to Automattic/wp-desktop-gh-bridge/pull/22.

Note: we're using a duplicate `mac` entry for now and will refactor later. In any case, we plan on deprecating the use of the bridge entirely in the near future, at which point the duplicated code will not be of concern.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -437,6 +437,72 @@ jobs:
                   set +e
                   rm -rf release/github
                   rm -rf release/mac
+      - persist_to_workspace:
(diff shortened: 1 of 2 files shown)


bsessions85:
LGTM! 🚢

PR #900, opened 2020-06-08
build: Add symlink to calypso .nvmrc to the root of the project

mokagio:
### Description:

`ln -s calypso/.nvmrc .nvmrc`

### Motivation and Context:

Every time I build the app for release, I get an error because my Node version is different from the Calypso one.

This should make it easier to keep the two versions in sync.

I don't think we should remove the version sync check in the build process, just in case.

### How Has This Been Tested:

- Set your system's Node version to something different from 12.16.2 (`cat ./calypso/.nvmrc`).
- `cd` into the project and run `yarn`
	- On `master`, it'll fail;
	- On this branch, it'll pass

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+calypso/.nvmrc


nsakaimbo:
This looks promising. I don't think I've added symlinks in Git before - this shows up as a new file in the diff, yeah?

How does this work if the submodule hasn't been initialized yet? (i.e. the calypso subfolder is missing to begin with). Will the symlink pick up the `.nvmrc` once added?

mokagio:
Sorry @nsakaimbo I somehow missed your comment. 😳 🙏 

Good question. In fact, it made me realize I didn't commit the symlinking properly. Unfortunately, now that the file is an actual symlink, there's a failure in CI.

Looks like I'll have more work to do here 😄 .

mokagio:
Now that I fixed the symlink, I can answer your question properly

> How does this work if the submodule hasn't been initialized yet? (i.e. the calypso subfolder is missing to begin with). Will the symlink pick up the `.nvmrc` once added?

Yes. Here's how it run on my end after a clean checkout

```
# ttt is the folder where I put the clean checkout
# note the node version is 13.12.0, the default for my system
ttt  develop [] is 📦 v6.0.0-beta2 via ⬢ v13.12.0
➜ git checkout --track origin/symlink-nvmrc
Branch 'symlink-nvmrc' set up to track remote branch 'symlink-nvmrc' from 'origin'.
Switched to a new branch 'symlink-nvmrc'

# the .nvmrc file is not here yet, because the sumbodule is not there, yet
ttt  symlink-nvmrc [] is 📦 v5.2.0 via ⬢ v13.12.0
➜ cat .nvmrc
[bat error]: '.nvmrc': No such file or directory (os error 2)

ttt  symlink-nvmrc [] is 📦 v5.2.0 via ⬢ v13.12.0
➜ g submodule init && g submodule update
Submodule 'wp-calypso' (https://github.com/Automattic/wp-calypso.git) registered for path 'calypso'
Cloning into '/Users/gio/Developer/a8c/ttt/calypso'...
Submodule path 'calypso': checked out '6891b7783184c36dfd3be93e7aabbba0cc099feb'

# now that the submodule has been fetched, the nvmrc is found
ttt  symlink-nvmrc [] is 📦 v5.2.0 via ⬢ v13.12.0 took 7m 31s
➜ cat .nvmrc
───────┬──────────────────────────────────────────────────────────────────────
       │ File: .nvmrc
───────┼──────────────────────────────────────────────────────────────────────
   1   │ 12.16.2
───────┴──────────────────────────────────────────────────────────────────────

# for an extra check here I go in and out of the folder to let nvm pick up the new version
# there might be a better way to do it
ttt  symlink-nvmrc [] is 📦 v5.2.0 via ⬢ v13.12.0
➜ .. && cd -
~/Developer/a8c/ttt
Found '/Users/gio/Developer/a8c/ttt/.nvmrc' with version <12.16.2>
Now using node v12.16.2 (npm v6.14.4)

# now the version is the one in the symlinked nvmrc 👋
ttt  symlink-nvmrc [] is 📦 v5.2.0 via ⬢ v12.16.2 took 2s
➜                                                     
```

I think it's ok if the `.nvmrc` is not available when we do a fresh checkout, because, after all, we'll need to fetch the submodules to do any kind of real work with the app anyways. What do you think?

nsakaimbo:
Closing - The migration of wp-desktop to the Calypso monorepo is almost final. This optimization is great but will be unnecessary going forward. Still - thanks for this change, @mokagio!

PR #904, opened 2020-06-08
ci: add Calypso PR author to Slack Status

nsakaimbo:
### Description

This PR adds Calypso PR author to CI status posted via Slack.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,6 +224,10 @@ references:
               }
             }
         }'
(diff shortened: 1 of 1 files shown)


bsessions85:
What about using something like `echo 'export CALYPSO_PR_AUTHOR="$(curl https://api.github.com/users/${pullRequestUserName} | jq --raw-output '.name')"' >> $BASH_ENV;` instead of the git request so we get the name of the user we are tagging?

PR #907, opened 2020-06-12
calypso: update to latest

nsakaimbo:
### Description

Update to latest calypso (Automattic/wp-calypso@931916f).

This calypso bump contains the following fixes that were made in upstream Calypso PRs:

* Fixed a bug where clicking "View Post" after updating/publishing a post broke the app (#908)
* Fixed a bug where the site setup list would not populate. To address this issue, _site setup list on desktop will be disabled_ until a more comprehensive solution is found. (#909)
* Fixed a bug where clicking "print recovery codes" would silently fail (#910)


nsakaimbo:
@mokagio I just found out the new Calypso support card was reverted on June 4th, and added back on June 9th. The Calypso SHA in 6.0.0-beta2 is as of June 8th (we just missed it!). Can we cut a beta3 release to get this update into our next release?

nsakaimbo:
I restored the `release/6.0.0` branch and am pointing the Calypso update there so all we need is to bump the version number once merged.

nsakaimbo:
Update: do not merge. We just learned of a regression in Calypso w.r.t the desktop app (the "Setup List" no longer works). Will try to get this fixed in Calypso so we can include the fix in the Calypso bump.

mokagio:
@nsakaimbo all right 👍 I'm subscribed to this. 👀

Issue #931, opened 2020-06-24
Domains link is not working from the My Sites > My Home > Get Help card


designsimply:
Steps to reproduce:

1. Go to My Sites > My Home.
1. Find the Get Help card.
1. Click on the "All about domains" link.

**Result:** a blank modal window pops up when the "All about domains" link in the My Sites > My Home > Get Help card is clicked from within the WPDesktop app. ([58s](https://cloudup.com/cKbbqeIdEPE))

**Note:** on the web, the "All about domains" link points to https://wordpress.com/support/domains/ and in the app it points to https://wordpress.com/support/all-about-domains/ and it appears the pop up window which displays the support link doesn't show search results if there are any in the same way the website does.

<img width="1280" alt="Screen Shot 2020-06-24 at Jun 24 3 01 54 PM" src="https://user-images.githubusercontent.com/1119271/85627852-6ce53b80-b62c-11ea-92fc-b181ce0d07e9.png">

<img width="1280" alt="Screen Shot 2020-06-25 at Jun 25 6 26 58 PM" src="https://user-images.githubusercontent.com/1119271/85808236-7ac0bc80-b711-11ea-8810-ef2f7ea1c7dd.png">

nsakaimbo:
Fixed by Automattic/wp-calypso#43543.

PR #933, opened 2020-06-25
calypso: update to Automattic/wp-calypso@6d53da1

nsakaimbo:
### Description

Update to Automattic/wp-calypso@6d53da1, which contains a fix for #931.


nsakaimbo:
> By the way, thanks for adding the link to the issue. For some reason, the commit link with only 7 character from the SHA doesn't work, it 404s.

huh - yeah not sure what happened, but will watch out for the formatting.

PR #939, opened 2020-07-14
ci: Add check to fail build if tag version doesn't match package.json

mokagio:
### Description

@nsakaimbo rightly noticed that when I pushed beta 1 for version 6.0.1, I used the wrong version. That is, `package.json` said `6.0.1` while the tag `v6.0.1-beta1`. Other than not being correct, this also causes issues with the auto-updater.

This PR adds a script to be run on CI when the build is from a tag that checks whether a given version matches the one from the `package.json`. If the versions don't match, the build will fail.

### How Has This Been Tested

I pushed a test tag and verified [the build doesn't get to the publish step](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/18849/workflows/adf57785-f02b-4bbf-94b2-922b87d4df0a/jobs/67292).

image

When the build [is not on a tag](https://app.circleci.com/pipelines/github/Automattic/wp-desktop/18850/workflows/0a99909d-9555-4b55-b907-ac402dbf37a5/jobs/67293), the validation is skipped.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -229,6 +229,26 @@ references:
         }'
 
 jobs:
+  validate_version:
+    docker:
+      - image: circleci/node:12.16.2-browsers
+    environment:
+      VERSION: << pipeline.git.tag >>
+    steps:
+      - when:
+          condition: << pipeline.git.tag >>
+          steps:
+            - checkout
+            - run:
+                name: Ensure package.json Version And Tag Match
+                command: node ./validate_tag.js $VERSION
+      - unless:
+          condition: << pipeline.git.tag >>
+          steps:
+            - run:
+                name: Skip Validation
+                command: echo "Not on a tag. Skipping version validation."
+
   build:
     docker:
       - image: circleci/node:12.16.2-browsers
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -229,6 +229,26 @@ references:
...
+            - checkout
+            - run:
+                name: Ensure package.json Version And Tag Match
+                command: node ./validate_tag.js $VERSION
+      - unless:
+          condition: << pipeline.git.tag >>
+          steps:
+            - run:
+                name: Skip Validation
+                command: echo "Not on a tag. Skipping version validation."
```
mokagio, line 250: This new step runs before the `build` one and if the build started with a tag, checks that the version in package.json matches the tag.

I'm not happy that we now have to run this step on every build, but at the same time it adds less that 15s to a build that is in the order of minutes. It seems like a reasonable tradeoff to prevent future error tagging new versions.

I thought about adding the condition in the `wp-desktop` workflow, but it already uses one and from [the docs](https://circleci.com/docs/reference-2-1/#the-when-attribute) it doesn't look like you can nest or compose them.

validate_tag.js
```diff
@@ -0,0 +1,30 @@
...
+const fs = require('fs');
+const package = JSON.parse(fs.readFileSync('./package.json', 'utf8'));
+const packageVersion = package['version'];
+
+if (package['version'] !== sanitizedVersion) {
+	console.error(`Expected version in package.json to match ${version}, got ${packageVersion}`);
+	process.exit(1);
+}
+
+console.log(`Version in package.json matches expected value ${version}. 👍`);
```
mokagio, line 30: Is the root of the project the best location for this file? I was tempted to put it into `.circleci` but I didn't want to "bind it" to CircleCI, it should be usable under any CI.

nsakaimbo: I think project root is totally fine here. 👍

PR #940, opened 2020-07-14
chore: Merge 6.0.1 beta 3

mokagio:


mokagio:
@nsakaimbo I forgot to mark this ready for review at the time of beta 2.

Will the change in 23c27df require a new beta? If so, maybe we should wait to merge this till the beta is ready?

nsakaimbo:
Hi, @mokagio! Thanks for checking. Was actually about to ping you about another beta deployment. I think we should deploy another beta version to verify the fixes that were recently committed. We can hold off on merging this until then. (Although I think if we merge-commit, not squash-merge, the change we can still continue working with this release branch - either way is fine.)

PR #944, opened 2020-07-17
fix: showMessageBox API usage (updater and SSO prompt)

nsakaimbo:
### Description

This PR fixes the native message boxes used to prompt the user (currently used for auto-update and SSO).

Turns out the `showMessageBox` API was updated in Electron v8 (undocumented breaking change) to return (1) a promise, and (2) a object `MessageBoxReturnValue` object (previously was a synchronous API that returned the index of the button tapped).

### DIFF
--- a/desktop/app-handlers/updater/auto-updater/index.js
+++ b/desktop/app-handlers/updater/auto-updater/index.js
@@ -85,6 +85,8 @@ class AutoUpdater extends Updater {
 	}
 
 	onConfirm() {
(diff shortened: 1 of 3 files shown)


nsakaimbo:
@rachelmcr Just a heads up this fixes the auto-update prompt and the SSO prompts which silently broke with the Electron upgrade. 

There are other regressions with respect to SSO that this particular change doesn't address (looking into those in separate PRs).

